PR #9, opened 2016-10-06
Fix getting container names of versions in openshift namespace; popd; done

omron93:
This PR fixes problem when version in openshift namespace  is not the lesser one (example of s2i-python-container). In this case, `NAMESPACE` variable is not set back to `openshift/` and created versions have wrong name.

@hhorak @bparees Same as sclorg/s2i-python-container#148. Please merge.


omron93:
[test]

@hhorak Please merge.

PR #22, opened 2017-05-24
Update Dockerfiles

omron93:
Provide same labels for all base image variants (CentOS, RHEL, Fedora)
    - add name, version, summary and description labels
    - remove architecture label for RHEL based images (this label is set in RHE$

(don't use separate instruction for each label)

@hhorak @praiskup @pkubatrh Please take a look and merge.


hhorak:
It looks good to me, I'd just suggest using a bit more verbose description, e.g. like in https://hhorak.fedorapeople.org/nginx-docker/Dockerfile.

omron93:
@hhorak Thanks for suggestion. It updated.

hhorak:
Sounds good to me, thanks, merging :)

PR #26, opened 2017-08-01
Move to newest common scripts

pkubatrh:


pkubatrh:
[test]

1.10/Dockerfile.rhel7
```diff
@@ -32,7 +32,10 @@ LABEL summary="$SUMMARY" \
 ENV NGINX_CONFIGURATION_PATH=/opt/app-root/etc/nginx.d
 ENV NGINX_DEFAULT_CONF_PATH=/opt/app-root/etc/nginx.default.d
 
-RUN yum install -y yum-utils && \
+# To use subscription inside container yum command has to be run first (before yum-config-manager)
+# https://access.redhat.com/solutions/1443553
+RUN yum repolist > /dev/null && \
```
omron93: @pkubatrh IMHO `yum repolist` is redundant there. Even without this, `yum` is run before `yum-config-manager` so everything should be fine.

pkubatrh: That is what I though as well. But the tests were failing before as they could not find rh-nginx* packages And it turned out it was because the rhscl repository was not getting enabled.

pkubatrh:
```
# yum search rh-nginx18
Warning: No matches found for: rh-nginx18
# yum-config-manager --enable rhel-server-rhscl-7-rpms
# yum search rh-nginx18
Warning: No matches found for: rh-nginx18
# yum repolist | grep rhscl
# yum-config-manager --enable rhel-server-rhscl-7-rpms
# yum repolist | grep rhscl
rhel-server-rhscl-7-rpms/7Server/x86_64 Red Hat Software Collections RPMs  8049
```

omron93: But this is really weird.

`yum install -y yum-utils` does nothing (yum-utils are already installed in s2i-base). But without this command (only with `yum repolist`) RHSCL repository isn't enabled too...

pkubatrh: One thing to note is that I am reproducing on a Workstation version of RHEL so I do not have `rhel-server-rhscl-7-rpms` enabled on the host.

omron93: I get same output on machine with subscription.
But I seems that even that the `yum install yum-utils` doesn't create anything, it updates repositories. So I don't know why it is not working.

Older `s2i-base` (without yum-utils) doesn't need `yum clean all` and first `yum search` finds rh-nginx18.

pkubatrh: In my case all that first `yum` command did was to populate the `redhat.repo` file which was until then empty. However, it seems that subsequent actions with that .repo file do not seem to affect anything other than the file itself. Which is why I though cleaning the cache might fix something but I have not expected the `yum clean` command to revert the .repo file to its previous state (`rhel-server-rhscl-7-rpms` got disabled)

pkubatrh: Just hit the same issue in mongodb: https://github.com/sclorg/mongodb-container/pull/253

pkubatrh: And I can replicate the issue the same way even on a plain `rhel7` image (7.4 based). So this might be a bug in the base image, or maybe in one of the yum packages.

pkubatrh: Created a bug against `yum-utils` to start with: https://bugzilla.redhat.com/show_bug.cgi?id=1479388
Let us see where this ends up...

pkubatrh:
The changes to RUN are now properly documented as a workaround with a reference to the bug I have created.

omron93:
@pkubatrh Thank you for going deeper in yum problem.

I think this can be merged now. And after reported yum bug is resolved a workaround can be removed. Thanks.

PR #27, opened 2017-08-20
Use s2i-core instead of s2i-base

hhorak:


torsava:
The changes look good, sorry for the delay!

PR #31, opened 2017-10-16
Add symlink to test-lib-openshift.sh in test dir.

pvalena:
The templates will reside in

https://github.com/sclorg/nginx-ex/tree/master/openshift/templates


pvalena:
[Changes Info](https://github.com/sclorg/s2i-python-container/pull/218#issuecomment-335276731)

hhorak:
[test]

PR #32, opened 2017-10-19
Pre-init s2i extensibility

### LINKED ISSUE #18: Support for environment variable replacement in the S2I configuration files

### LINKED ISSUE #20: support for modification to global config via user supplied sed script
The [httpd image has support](https://github.com/sclorg/httpd-container/blob/master/2.4/s2i/bin/assemble) for developer provided sed input that can override the global config.

Can the nginx image be altered so a similar mechanism can be used to do things like enable `/nginx_status` with the `server { }` block?

kosciCZ:
Hi,

This PR would add ability to execute user supplied scripts just before nginx start. This would fix #18, because it implements #19, and it would also fix #20.

All user has to do is place `.sh` scripts in `./nginx-pre-init/` folder, and these scripts will then be sourced right before nginx starts.

I have also modified permissions on config files, for when container is not running on user 1001 (as seen in tests)

This PR also includes test case for each version and modified test running script to be able to run more than one test. 

I've separated changes into three commits for each version to track changes more transparently. Added functionality is the same in all of them, except for 1.8, where I decided to add option to extend the default server block, which I took from 1.10 & 1.12.

### DIFF
--- a/1.10/s2i/bin/assemble
+++ b/1.10/s2i/bin/assemble
@@ -31,7 +31,5 @@ if [ -d ./nginx-pre-init ]; then
   fi
 fi
 
-mkdir -p ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-pre-init/
-
 # Fix source directory permissions
 fix-permissions ./
--- a/1.10/s2i/bin/assemble
+++ b/1.10/s2i/bin/assemble
@@ -11,6 +11,7 @@ if [ -d ./nginx-cfg ]; then
     cp -v ./nginx-cfg/*.conf "${NGINX_CONFIGURATION_PATH}"
     rm -rf ./nginx-cfg
   fi
+  chmod -Rf g+rw ${NGINX_CONFIGURATION_PATH}
 fi
 
 if [ -d ./nginx-default-cfg ]; then
@@ -19,7 +20,18 @@ if [ -d ./nginx-default-cfg ]; then
     cp -v ./nginx-default-cfg/*.conf "${NGINX_DEFAULT_CONF_PATH}"
     rm -rf ./nginx-default-cfg
   fi
+  chmod -Rf g+rw ${NGINX_DEFAULT_CONF_PATH}
 fi
+
+if [ -d ./nginx-pre-init ]; then
+  echo "---> Copying nginx pre-init scripts..."
+  if [ "$(ls -A ./nginx-pre-init/*)" ]; then
+    cp -v ./nginx-pre-init/* "${NGINX_APP_ROOT}/etc/nginx-pre-init"
+    rm -rf ./nginx-pre-init
+  fi
+fi
+
+mkdir -p ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-pre-init/
 
 # Fix source directory permissions
 fix-permissions ./
--- a/1.10/test/run
+++ b/1.10/test/run
@@ -32,7 +32,7 @@ container_ip() {
 }
 
 run_s2i_build() {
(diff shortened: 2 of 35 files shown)


hhorak:
[test]

kosciCZ:
@hhorak I have added a fix, can you please trigger the tests again?

hhorak:
[test]

kosciCZ:
ping, can someone please review this PR?

1.10/test/run
```diff
@@ -174,44 +188,83 @@ test_connection() {
...
-  check_result $?
+  check_result $? "test-app"
 
   test_connection
-  check_result $?
+  check_result $? "test-app"
   cleanup_test_app
 }
 
+test_pre_init_script() {
```
omron93: Definitely +1 for passing app name as argument to functions. So be able to use these functions in more cases.

`test_pre_init_script` and `test_application`
`test_connection_s2i` and `test_connection`
are very similar. Only functions used for s2i app testing adds 1-2 lines of code.

I suggest:
    1. remove `cleanup_test_app` from `test_application` and have to call it explicitly after testing of each app finished.
    (2. rename `test_application` to `test_application_basics`)
    3. when testing s2i app use same `test_application` and before calling `cleanup_test_app` do

```
test_command "cat /opt/app-root/etc/nginx.d/default.conf | grep 'resolver' | sed -e 's/resolver\(.*\);/\1/' | grep 'DNS_SERVER'" ""
check_result $? "pre-init-test-app"

test_for_output "http://$(container_ip):${test_port}/aliased/index2.html" "NGINX2 is working" 
```

1.10/s2i/bin/assemble
```diff
@@ -19,6 +20,15 @@ if [ -d ./nginx-default-cfg ]; then
     cp -v ./nginx-default-cfg/*.conf "${NGINX_DEFAULT_CONF_PATH}"
     rm -rf ./nginx-default-cfg
   fi
+  chmod -Rf g+rw ${NGINX_DEFAULT_CONF_PATH}
+fi
+
+if [ -d ./nginx-pre-init ]; then
+  echo "---> Copying nginx pre-init scripts..."
+  if [ "$(ls -A ./nginx-pre-init/*)" ]; then
+    cp -v ./nginx-pre-init/* "${NGINX_APP_ROOT}/etc/nginx-pre-init"
```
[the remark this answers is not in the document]
kosciCZ: Yeah, that makes more sense. I've used this location before, don't know why I chose something else here than `NGINX_CONTAINER_SCRIPTS_PATH`. Will change that

1.10/test/run
kosciCZ: +1, will implement this. Just a note, it's the other way around, `test_connection` has one more line to test. But i got your point, and I'll fix my code.

1.10/s2i/bin/assemble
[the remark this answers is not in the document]
hhorak: I've changed this in #41, since I'm not allowed to push to this branch.

1.10/test/run
hhorak: I haven't changed this in #41, because I'd rather see using the `test.lib.sh` from common repo.

hhorak:
I wanted to push changes into this PR directly, but it didn't work, so I opened a new PR #41, using most of the changes here, but also doing slightly more.

hhorak:
PR #41 was merged, closing this one. Anyway, thanks a lot for large portion of work, @kosciCZ.

PR #39, opened 2017-12-14
Do not use VOLUMES because it fails in OpenShift Online

hhorak:
It's rationalized in https://github.com/sclorg/httpd-container/issues/30

Ideally this should be reverted once OpenShift Online is fixed.


hhorak:
Review of the same change already done in https://github.com/sclorg/httpd-container/pull/41.

PR #41, opened 2017-12-16
Slightly changed S2i extensible from kosciCZ

hhorak:
This is an implementation of flexible feature for the container. It is almost identical what @kosciCZ did in #32. It only puts extended files into place suggested during PR #32 review, and then `pre-init` hook name was changed to `start-hook`, which was also chosen hook name in https://github.com/sclorg/postgresql-container/pull/208.


hhorak:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
Most of these changes were already reviewed as part of PR #32.
FYI @uhliarik, can you provide final ack if you're fine with these changes?

hhorak:
[test-openshift]

PR #42, opened 2017-12-18
Couple updates from container-images/nginx repo.

phracek:

This is Pull Request from container-images/nginx repo.

It updates mainly LABELS.

### DIFF
--- a/1.10/Dockerfile
+++ b/1.10/Dockerfile
@@ -8,7 +8,12 @@ FROM centos/s2i-core-centos7
 EXPOSE 8080
 EXPOSE 8443
 
-ENV NGINX_VERSION=1.10
+ENV NAME=nginx \
+    NGINX_VERSION=1.10 \
+    NGINX_SHORT_VER=110 \
+    VERSION=0 \
+    RELEASE=1 \
+    ARCH=x86_64
 
 ENV SUMMARY="Platform for running nginx $NGINX_VERSION or building nginx-based application" \
     DESCRIPTION="Nginx is a web server and a reverse proxy server for HTTP, SMTP, POP3 and IMAP \
@@ -20,14 +25,16 @@ Nginx server image can be extended using source-to-image tool."
 LABEL summary="$SUMMARY" \
       description="$DESCRIPTION" \
       io.k8s.description="$DESCRIPTION" \
-      io.k8s.display-name="Nginx 1.10" \
+      io.k8s.display-name="Nginx $NGINX_VERSION" \
       io.openshift.expose-services="8080:http" \
       io.openshift.expose-services="8443:https" \
-      io.openshift.tags="builder,nginx,rh-nginx110" \
-      com.redhat.component="rh-nginx110-docker" \
-      name="centos/nginx-110-centos7" \
-      version="1.10" \
-      release="1"
+      io.openshift.tags="builder,$NAME,rh-$NAME$NGINX_SHORT_VER" \
+      com.redhat.component="rh-$NAME$NGINX_SHORT_VER-docker" \
+      name="centos/$NAME-$NGINX_SHORT_VER-centos7" \
+      version="$NGINX_VERSION" \
+      release="$RELEASE" \
+      architecture="$ARCH" \
+      maintainer="SoftwareCollections.org <sclorg@redhat.com>"
 
 ENV NGINX_CONFIGURATION_PATH=/opt/app-root/etc/nginx.d
 ENV NGINX_DEFAULT_CONF_PATH=/opt/app-root/etc/nginx.default.d
--- a/1.10/Dockerfile
+++ b/1.10/Dockerfile
@@ -12,8 +12,6 @@ ENV NAME=nginx \
     NGINX_VERSION=1.10 \
     NGINX_SHORT_VER=110 \
     VERSION=0 \
-    RELEASE=1 \
-    ARCH=x86_64
 
 ENV SUMMARY="Platform for running nginx $NGINX_VERSION or building nginx-based application" \
     DESCRIPTION="Nginx is a web server and a reverse proxy server for HTTP, SMTP, POP3 and IMAP \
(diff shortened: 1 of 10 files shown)


1.10/Dockerfile
```diff
@@ -8,7 +8,12 @@ FROM centos/s2i-core-centos7
 EXPOSE 8080
 EXPOSE 8443
 
-ENV NGINX_VERSION=1.10
+ENV NAME=nginx \
+    NGINX_VERSION=1.10 \
+    NGINX_SHORT_VER=110 \
+    VERSION=0 \
+    RELEASE=1 \
+    ARCH=x86_64
```
hhorak: Hard-coding arch is not a good idea, it would make us trouble once there are multi-arch containers.

1.10/Dockerfile
```diff
@@ -20,14 +25,16 @@ Nginx server image can be extended using source-to-image tool."
...
-      com.redhat.component="rh-nginx110-docker" \
-      name="centos/nginx-110-centos7" \
-      version="1.10" \
-      release="1"
+      io.openshift.tags="builder,$NAME,rh-$NAME$NGINX_SHORT_VER" \
+      com.redhat.component="rh-$NAME$NGINX_SHORT_VER-docker" \
+      name="centos/$NAME-$NGINX_SHORT_VER-centos7" \
+      version="$NGINX_VERSION" \
+      release="$RELEASE" \
+      architecture="$ARCH" \
```
hhorak: We don't actually need to define architecture for RHEL and CentOS images, please, remove it, since it would only make us troubles once containers work on other arches.

1.10/README.md
```diff
@@ -62,6 +62,12 @@ The nginx container image supports the following configuration variable, which c
 | :--------------------- | ----------------------------------------- |
 |  `NGINX_LOG_TO_VOLUME` | When `NGINX_LOG_TO_VOLUME` is set, nginx logs into `/var/opt/rh/rh-nginx110/log/nginx/` |
 
+t your own web root like this:
+```
+$ docker run -v <DIR>:/var/www/html/
```
hhorak: It doesn't seem like this is a working example, there is no image name for example. Also the line above "t your own web root like this:" looks like incomplete.

hhorak:
@phracek Thank you for the submission. Although there are some issues mentioned in-line.

hhorak:
Sorry for closing/re-opening, I clicked the wrong button.

omron93:
@phracek Please take a look to https://github.com/sclorg/s2i-php-container/pull/187.

@pkubatrh I guess curly braces should be also in nginx-container Dockerfiles, right?

1.10/Dockerfile
phracek: Removed.

phracek:
See labels in Fedora Container Guidelines https://fedoraproject.org/wiki/Container:Guidelines#LABELS.
I guess, if it is a bug, it should be fixed also on Guidelines. 

Unfortunately, also http://docs.projectatomic.io/container-best-practices/#_labels does not provide this kind of information. Can you please file an issue on it to container-best-practices? I think this is important and we have to standardize.

pkubatrh:
>@pkubatrh I guess curly braces should be also in nginx-container Dockerfiles, right?

Yes, I expect it would cause failures here as well. I am still looking into this, I will open up some bug reports once I know what this is caused by.

phracek:
All nginx Dockerfiles contains curly braces.

pkubatrh:
>All nginx Dockerfiles contains curly braces.

Missing in 1.8 versions

1.10/Dockerfile
omron93: @phracek So no need to have ARCH env var.

Also we are removing `release` labels from CentOS/RHEL images.

pkubatrh:
Submitted a PR against dockerfile-parser to fix the variable concatenation issue: https://github.com/DBuildService/dockerfile-parse/pull/45

1.10/Dockerfile
phracek: Why? https://fedoraproject.org/wiki/Container:Guidelines#LABELS contains release label as required. Although, it will be still 1.

pkubatrh: The way I see it the Fedora version of release label also has a DISTTAG suffix added (that expands to something like f26container) which is why the release label has to be present. We do not have anything like that in RHEL/CentOS so we can just remove it completely and let the build system increment this.

Also the guidelines say that the release shout NOT stay at 1:
"release: Populated from the ENV variable. Should start at 1, and increment every time the image changes."

phracek: Increasing RELEASE should be done by each change. Yes, Your team should increase it. I will do it over all my changes in nginx container.

hhorak: Increasing the release in upstream sources is really not necessary, because here on github we generally don't know how many times we needed to increase it downstream before and there is no reason to keep it in sync. If the release is necessary for Fedora Dockerfile, fine, let's keep it there for the sake of having the line on a correct place, but in Dockerfiles for other distros, it's really not necessary now, because internal OSBS bumps it automatically and we don't use it at all for CentOS images.

hhorak:
[test]

1.10/Dockerfile
```diff
@@ -8,7 +8,10 @@ FROM centos/s2i-core-centos7
 EXPOSE 8080
 EXPOSE 8443
 
-ENV NGINX_VERSION=1.10
+ENV NAME=nginx \
+    NGINX_VERSION=1.10 \
+    NGINX_SHORT_VER=110 \
+    VERSION=0 \
```
hhorak: @phracek the backslash here needs to be removed, build fails because of this.

phracek: sorry for stupid typo. Fixed.

omron93:
[test]

omron93:
LGTM. @phracek Thanks for contribution.

PR #45, opened 2018-01-29
enable substituting nginx.conf

TomasTomecek:
Please suggest how should I test this feature.

### DIFF
--- a/1.12/Dockerfile.fedora
+++ b/1.12/Dockerfile.fedora
@@ -37,8 +37,9 @@ LABEL summary="${SUMMARY}" \
       help="For more information visit https://github.com/sclorg/${NAME}-container" \
       usage="s2i build file:///your/app ${FGC}/nginx your-app"
 
-ENV NGINX_CONFIGURATION_PATH=/opt/app-root/etc/nginx.d
-ENV NGINX_DEFAULT_CONF_PATH=/opt/app-root/etc/nginx.default.d
+ENV NGINX_CONFIGURATION_PATH=/opt/app-root/etc/nginx.d \
+    NGINX_CONF_PATH=/etc/nginx/nginx.conf \
+    NGINX_DEFAULT_CONF_PATH=/opt/app-root/etc/nginx.default.d
 
 RUN dnf install -y gettext hostname && \
     INSTALL_PKGS="nss_wrapper bind-utils nginx" && \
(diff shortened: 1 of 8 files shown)


omron93:
To run test of this new app, `$VERSION/test/run` file has to be modified.

TomasTomecek:
Should be correct now. Sorry it took so long.

Once you ack the design, I'll squash the commits (to two, since the other commit actually fixes the 1.12 fedora dockerfile).

1.12/test/run
```diff
@@ -236,10 +236,18 @@ cid_file=$(mktemp -u --suffix=.cid)
 
 # Since we built the candidate image locally, we don't want S2I attempt to pull
 # it from Docker hub
-s2i_args="--pull-policy=never"
+s2i_build_args="--pull-policy=never --assemble-user=root"
```
omron93: `root` isn't possible in OpenShift. So I'm not sure if we wants this...

1.12/Dockerfile.fedora
```diff
@@ -58,10 +60,13 @@ COPY ./root/ /
 RUN sed -i -f /opt/app-root/nginxconf-fed.sed /etc/nginx/nginx.conf && \
     mkdir -p /opt/app-root/etc/nginx.d/ && \
     mkdir -p /opt/app-root/etc/nginx.default.d/ && \
+    mkdir -p ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
     chmod -R a+rwx /opt/app-root/etc && \
     chmod -R a+rwx /var /run && \
+    chmod -R a+rwx ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
```
omron93: Why 'a+rwx'? Permissions are too loosen.

TomasTomecek: I just copied from different line w/o thinking too much about it.

1.12/test/run
TomasTomecek: I kind of expected this. So can we put the file in app root and just symlink?

1.12/Dockerfile.fedora
omron93: Container should be able to run with default user and with randum UID+root group. `nginx-start` is owned by root group, ug+rwx should be enought there...

1.12/test/run
omron93: > . So can we put the file in app root and just symlink?

That's one possible solution. Or we should be able to rewrite it, am I right?

TomasTomecek: Yup, that sounds good as well: just let me know which solution I should do.

omron93: It's up to you.

I prefer to rewrite file. But I'm not aware of any hard reason for this, so it's totally up to you.

TomasTomecek:
I fixed those two issues (the first one via `chmod a+rwx ${NGINX_CONF_PATH}`) and squashed commits so this PR is ready to be merged from my PoV.

omron93:
Thanks.
[test][test-openshift]

PR #48, opened 2018-02-07
Rename nginx-start-hook extending point to nginx-start

hhorak:
We would like to have the same naming convention across images for extending feature (flexible image). Since the extendibility is quite fresh feature in the nginx image, it is fine to expect that not many users will use it and especially that they'd use `nging-start-hook`. In case we get feedback about breaking compatibility, we can always add the name `nging-start-hook` back and work with both locations.


hhorak:
[test]
[test-openshift]

hhorak:
[test]
[test-openshift]

1.10/s2i/bin/assemble
```diff
@@ -23,11 +23,11 @@ if [ -d ./nginx-default-cfg ]; then
   chmod -Rf g+rw ${NGINX_DEFAULT_CONF_PATH}
 fi
 
-if [ -d ./nginx-start-hook ]; then
+if [ -d ./nginx-start ]; then
   echo "---> Copying nginx start-hook scripts..."
```
omron93: I'm thinking about `start-hook scrips` if it's misleading or not... Thought?

hhorak: I agree that it would be maybe a bit more precise, but what we really want to achieve here is to have the same names used in more images. And the `-hook` suffix is not used in other images, that's why it is prioritised. In the end, what is more important is a documentation of how the concrete hook is described.

omron93:
LGTM. Merging.

PR #56, opened 2018-06-05
Update context-dir path

CountPickering:
When running an s2i build the following error message occurred:

> error: build error: provided context directory does not exist: examples/1.12/test/test-app

Removed the "test" folder from the path to reflect the current location of test-app.


omron93:
@CountPickering You are right. Thanks for the fix.

Docs only change, so no testing is needed. Merging.

CountPickering:
Happy to help!

PR #57, opened 2018-06-05
Use released repos only

hhorak:
It is not necessary to use testing repo now anymore, all packages are in released repo already.


omron93:
Images build fine. Thanks.

PR #59, opened 2018-06-08
Fix stdout by follow upstream logging practices for nginx in docker

### LINKED ISSUE #53: access logging not going to stdout
The [nginxconf.sd](https://github.com/sclorg/nginx-container/blob/master/1.12/root/opt/app-root/nginxconf.sed#L7) script is removing the access log line rather then redirecting logging to stdout like the error log.

Currently:

```
s%/var/opt/rh/rh-nginx112/log/nginx/error.log%stderr%
s%access_log  /var/opt/rh/rh-nginx112/log/nginx/access.log  main;%%
```

I would expect instead:

```
s%/var/opt/rh/rh-nginx112/log/nginx/error.log%stderr%
s%/var/opt/rh/rh-nginx112/log/nginx/access.log%stdout%
```

gregswift:
Per upstream nginx images, just symlink the default log files to the stdout/err devices and dont bother trying to reconfigure the nginx.conf file

https://stackoverflow.com/questions/22541333/have-nginx-access-log-and-error-log-log-to-stdout-and-stderr-of-master-process/42369571#42369571
https://github.com/nginxinc/docker-nginx/issues/87
https://github.com/nginxinc/docker-nginx/blob/master/mainline/alpine/Dockerfile#L136-L137

Closes #53

### DIFF
--- a/1.10/Dockerfile
+++ b/1.10/Dockerfile
@@ -55,16 +55,19 @@ COPY ./root/ /
 # In order to drop the root user, we have to make some directories world
 # writeable as OpenShift default security model is to run the container under
 # random UID.
-RUN sed -i -f ${NGINX_APP_ROOT}/nginxconf.sed /etc/opt/rh/rh-nginx110/nginx/nginx.conf && \
+RUN sed -i -f ${NGINX_APP_ROOT}/nginxconf.sed /etc/opt/rh/rh-nginx${NGINX_SHORT_VER}/nginx/nginx.conf && \
     mkdir -p ${NGINX_APP_ROOT}/etc/nginx.d/ && \
     mkdir -p ${NGINX_APP_ROOT}/etc/nginx.default.d/ && \
     mkdir -p ${NGINX_APP_ROOT}/src/nginx-start/ && \
     mkdir -p ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
+    mkdir -p /var/opt/rh/rh-nginx${NGINX_SHORT_VER}/log/nginx && \
+    ln -sf /dev/stdout /var/opt/rh/rh-nginx${NGINX_SHORT_VER}/log/nginx/access.log && \
+    ln -sf /dev/stderr /var/opt/rh/rh-nginx${NGINX_SHORT_VER}/log/nginx/error.log && \
     chmod -R a+rwx ${NGINX_APP_ROOT}/etc && \
-    chmod -R a+rwx /var/opt/rh/rh-nginx110 && \
+    chmod -R a+rwx /var/opt/rh/rh-nginx${NGINX_SHORT_VER} && \
     chmod -R a+rwx ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
     chown -R 1001:0 ${NGINX_APP_ROOT} && \
-    chown -R 1001:0 /var/opt/rh/rh-nginx110 && \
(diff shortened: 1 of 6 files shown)


omron93:
Thanks for the fix. Using symlinks looks good.

@hhorak @uhliarik Do you remember what was a reason to remove access_log line? (to hide user names and passwords or something like this... ?)

hhorak:
> @hhorak @uhliarik Do you remember what was a reason to remove access_log line? (to hide user names and passwords or something like this... ?)

git blame tells me that it was done in very early state by @hanzz -- Honzo, do you remember by any chance?

I'd like to hear what @notroj or @uhliarik think, because the log can include some sensitive information that might be sometimes part of the url (tokens, hashes).. Maybe we can redirect access.log to stdout only when some variable is specified and by default not change the default?

omron93:
> Maybe we can redirect access.log to stdout only when some variable is specified and by default not change the default?

Good to know what is mentioned in issue #53 

`Tried to make this work and found that the access_log match just doesnt do anything (at least i nthe 1.10 branch) because there is no access_log line for sed to match.`

gregswift:
A couple of additional notes:

a) I didn't update any docs because the (docs already say stdout works)[https://github.com/sclorg/nginx-container/tree/master/1.10#troubleshooting]

```
By default, nginx logs into standard output, so the log is available in the container log
```

b) Because there is no access_logs line defined a user already has 2 ways to change the behavior.

1: (mount a volume )[https://github.com/sclorg/nginx-container/tree/master/1.10#environment-variables-and-volumes] although with this method they dont need a variable to accomplish it, just the mount so that could use a document update.

2: adding a file to their git repo in `./nginx-cfg/*.conf` that defined how they want the access_logs to be handled (I havent tested this but it should work)

1.10/Dockerfile
```diff
@@ -60,6 +60,9 @@ RUN sed -i -f ${NGINX_APP_ROOT}/nginxconf.sed /etc/opt/rh/rh-nginx110/nginx/ngin
     mkdir -p ${NGINX_APP_ROOT}/etc/nginx.default.d/ && \
     mkdir -p ${NGINX_APP_ROOT}/src/nginx-start/ && \
     mkdir -p ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
+    mkdir -p /var/opt/rh/rh-nginx110/logs/nginx && \
+    ln -sf /dev/stdout /var/opt/rh/rh-nginx110/log/nginx/access.log && \
```
kamalgill: Should this line (and the previous/next) ref `NGINX_SHORT_VER` instead of hard-coding the version in `.../rh-nginx110/...`?

gregswift: I changed it to use that var, and then extended it to several other lines... *but* there are several i didn't touch.  Feels like it more should be converted? whats the distinction between doing it or not?

gregswift:
Whats the next step after approval to get this merged ?

kamalgill:
Good question @gregswift .  The changes look good to me, but unfortunately I don't have permissions to merge this PR.

omron93:
@hanzz @notroj @uhliarik Please take a look into this PR?

uhliarik:
Shouldn't path in mkdir contain directory "log" instead of "logs"?

rackergs:
good catch @uhliarik .. fixed

uhliarik:
+1. LGTM.

omron93:
[test][test-openshift]

omron93:
Tests are passing.

@uhliarik Thanks for review. Merging.

PR #73, opened 2019-01-03
Add Nginx 1.14 container

patrick-tavares:
As the title implies, this PR adds an Nginx 1.14 container build and S2I capability. In addition, I made some changes to the Fedora image handling to properly install the specified version of Nginx (in this case 1.14) regardless of which version of Fedora the image is built against (this assumes that the version of Fedora still contains that specific version of Nginx in its repo).

I was able to test both the Centos and Fedora builds but not the RHEL build (don't have access to an entitled RHEL build box).

### DIFF
--- a/1.14/Dockerfile
+++ b/1.14/Dockerfile
@@ -1,0 +1,87 @@
+FROM centos/s2i-core-centos7
+
+# RHSCL rh-nginx114 image.
+#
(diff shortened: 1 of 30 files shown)


patrick-tavares:
Just saw PR #71. I should have checked that first. Closing.

PR #78, opened 2019-03-19
Sync 1.10 with 1.12/1.14.

notroj:
SSIA


notroj:
[test] [test-openshift]

PR #80, opened 2019-05-01
fix sed script and error_log path for Fedora

kubco2:


kubco2:
[test-openshift]

kubco2:
[test]

notroj:
Thanks!  Tests are broken for EOL Fedora, not sure why we are still running those.

PR #83, opened 2019-05-10
Merge nginx.conf file for different distros for 1.14

### LINKED ISSUE #82: rhel8 s2i build uses wrong path in nginx.conf
The instructions on https://access.redhat.com/containers/?tab=overview&get-method=red-hat-login#/registry.access.redhat.com/rhel8/nginx-114 using 1.14 produce a s2i container that doesn't start:
```
$ docker run --rm -it nginx-sample-app:latest
2019/05/07 21:52:37 [emerg] 1#0: open() "/etc/opt/rh/rh-nginx114/nginx/mime.types" failed (2: No such file or directory) in /etc/nginx/nginx.conf:28
nginx: [emerg] open() "/etc/opt/rh/rh-nginx114/nginx/mime.types" failed (2: No such file or directory) in /etc/nginx/nginx.conf:28
```

The paths in rhel8 have been changed, specifically mime.types:
```
$ docker run --rm -it nginx-sample-app:latest rpm -qf /etc/nginx/mime.types
nginx-1.14.1-8.module+el8+2505+fe936cef.x86_64
```

hhorak:
I'm not entirely sure this is a correct thing to do, so take it more as a WIP to fix #82.

With the current design of the testing app, users (or a test script) need to shuffle with config files to make the example app working, which is not acceptable user experience.

This change makes only one copy of the nginx.conf config file, while using the non-SCL paths. For making it work, we need to add compatibility symlinks for directories used in this config file.

Particularly, /usr/share/nginx and /etc/nginx will point to respective SCL directories, and /var/run needs to be writable the same as in RHEL-8 case.


hhorak:
I've thought about it a bit more, and it looks like the correct thing to do still. I'd like to see some feedback from anybody else though, especially whether somebody sees a problem with the symlinks.

hhorak:
The same changes are done in 1.12 version as well now.

hhorak:
@uhliarik what do you think?

hhorak:
[test-openshift]

uhliarik:
LGTM, this should work back-ward compatible as you mentioned above.

uhliarik:
[test-openshift]

PR #85, opened 2019-06-12
Replace docker cp with docker exec <cont> cat

hhorak:
When running with podman older than 1.1.0, which does not have cp command,
the test failed. This way the test does pretty much the same thing, but allows
to run it successfully with older podman as well.


notroj:
Nice, thanks, this is actually simpler too.  I'll merge if once tests pass.

PR #91, opened 2019-07-31
Fix persmissions for all files given in s2i

hhorak:
Otherwise we can see permissions denied in cases the files copied into
/usr/share did not have correct permissions on the host originaly.

It looked like this in the test log:
/usr/share/container-scripts/nginx/common.sh: line 33: /usr/share/container-scripts/nginx/nginx-start/init.sh: Permission denied

### DIFF
--- a/1.10/s2i/bin/assemble
+++ b/1.10/s2i/bin/assemble
@@ -3,7 +3,7 @@
 set -e
 
 echo "---> Installing application source"
-cp -Rf /tmp/src/. ./
+mv /tmp/src/* ./
 
 # Fix source directory permissions
 fix-permissions ./
@@ -11,7 +11,7 @@ fix-permissions ./
 if [ -d ./nginx-cfg ]; then
   echo "---> Copying nginx configuration files..."
   if [ "$(ls -A ./nginx-cfg/*.conf)" ]; then
-    cp -v ./nginx-cfg/*.conf "${NGINX_CONFIGURATION_PATH}"
+    cp -av ./nginx-cfg/*.conf "${NGINX_CONFIGURATION_PATH}"
     rm -rf ./nginx-cfg
   fi
   chmod -Rf g+rw ${NGINX_CONFIGURATION_PATH}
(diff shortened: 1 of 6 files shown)


hhorak:
@kubco2 This is the correct fix IMO.

1.10/s2i/bin/assemble
```diff
@@ -3,12 +3,15 @@
 set -e
 
 echo "---> Installing application source"
-cp -Rf /tmp/src/. ./
+mv /tmp/src/* ./
```
pkubatrh: Can we leave the `cp` for now, until we choose what to do about this for all images?

Similarly to other versions

phracek:
```suggestion
cp -Rf /tmp/src/. ./
```

hhorak: Thx.

1.12/s2i/bin/assemble
```diff
@@ -3,7 +3,10 @@
 set -e
 
 echo "---> Installing application source"
-cp -Rf /tmp/src/. ./
+mv /tmp/src/* ./
```
phracek:
```suggestion
cp -Rf /tmp/src/. ./
```

1.14/s2i/bin/assemble
```diff
@@ -3,7 +3,10 @@
 set -e
 
 echo "---> Installing application source"
-cp -Rf /tmp/src/. ./
+mv /tmp/src/* ./
```
phracek:
```suggestion
cp -Rf /tmp/src/* ./
```

phracek:
As soon as `mv` is replaced with `cp` then we can merge it.

1.14/s2i/bin/assemble
```diff
@@ -3,7 +3,10 @@
 set -e
 
 echo "---> Installing application source"
-cp -Rf /tmp/src/. ./
+cp -Rf /tmp/src/* ./
```
phracek:
```suggestion
cp -Rf /tmp/src/. ./
```

phracek: It was my fault sorry.

phracek:
Fedora test is failing on version.

```bash
17:03:05 [1m[INFO] Testing the image SCL enable...[0m
17:03:05 
17:03:06 ERROR[/bin/bash -c nginx -v] Expected 'nginx version: nginx/1.14.', got 'nginx version: nginx/1.16.1'
17:03:06 
17:03:06 [1m[INFO] TEST FAILED (1)...[0m
17:03:06 
17:03:06 
1
```

hhorak:
I've fixed that, Fedora dockerfiles indeed installed wrong version, but since we have `nginx` modules in Fedora now, we can install a correct version. We should be ok to merge now.

PR #93, opened 2019-09-23
Resolves: #1753002 - RFE: Add rh-nginx112-mod-stream to the rhscl/nginx-112-rhel7 image 

uhliarik:
Install rh-nginx112-nginx-mod-stream


hhorak:
The bz reference should ideally be in git commit, as nobody will find this pull-request when looking for a reason why this package was added.

pkubatrh:
lgtm

PR #97, opened 2019-11-22
add rhel8-based nginx-116 image

pkubatrh:


pkubatrh:
[test]

1.16/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+ENV NGINX_CONFIGURATION_PATH=${APP_ROOT}/etc/nginx.d \
+    NGINX_CONF_PATH=/etc/opt/rh/rh-nginx${NGINX_SHORT_VER}/nginx/nginx.conf \
+    NGINX_DEFAULT_CONF_PATH=${APP_ROOT}/etc/nginx.default.d \
+    NGINX_CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/nginx \
+    NGINX_APP_ROOT=${APP_ROOT} \
+    NGINX_LOG_PATH=/var/opt/rh/rh-nginx${NGINX_SHORT_VER}/log/nginx
+
+RUN yum install -y yum-utils gettext hostname && \
+    yum install -y centos-release-scl-rh && \
```
hhorak: https://cbs.centos.org/repos/sclo7-rh-nginx116-rh-candidate/x86_64/os/ can be used meanwhile, till the packages are available on the mirror.

1.16/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+# RHSCL rh-nginx114 image.
+#
+# Volumes:
+#  * /var/opt/rh/rh-nginx114/log/nginx/ - Storage for logs
+
+EXPOSE 8080
+EXPOSE 8443
+
+ENV NAME=nginx \
+    NGINX_VERSION=1.14 \
```
hhorak: I'd rather use correct versions and if the packages are not available yet, let the container build fail.

1.16/Dockerfile.fedora
```diff
@@ -0,0 +1,83 @@
...
+# nginx 1.14 image.
+#
+# Volumes:
+#  * /var/log/nginx/ - Storage for logs
+
+EXPOSE 8080
+EXPOSE 8443
+
+ENV NAME=nginx \
+    NGINX_VERSION=1.14 \
```
hhorak: nginx1.16 actually is in f29 already, so the dockerfile can be changed to 1.16, but it would be also good to use non-EOL Fedora version, not sure the updated s2i images for f30/f31are already in stable though.

1.16/Dockerfile.rhel7
```diff
@@ -0,0 +1,93 @@
...
+# RHSCL rh-nginx114 image.
+#
+# Volumes:
+#  * /var/opt/rh/rh-nginx114/log/nginx/ - Storage for logs
+
+EXPOSE 8080
+EXPOSE 8443
+
+ENV NAME=nginx \
+    NGINX_VERSION=1.14 \
```
hhorak: RHSCL 3.4 beta is already available, so we can definitely have 1.16 here already. Beta channel can be used below to make it work fine now, or we can simply keep the Dockerfile with main RHSCL channel as it is now and let the users who need to build it themselves with beta packages to change it (there are not many folks who do it I guess).

1.16/s2i/bin/usage
```diff
@@ -0,0 +1,17 @@
...
+NAMESPACE=centos
+[[ $DISTRO =~ rhel* ]] && NAMESPACE=rhscl
+
+cat <<EOF
+This is a S2I ${IMAGE_DESCRIPTION} ${DISTRO} base image:
+To use it, install S2I: https://github.com/openshift/source-to-image
+
+Sample invocation:
+
+s2i build https://github.com/sclorg/nginx-container.git --context-dir=1.16/test/test-app/ ${NAMESPACE}/nginx-116-${DISTRO}7 nginx-sample-app
```
hhorak: `$NGINX_VERSION` can be used here, it will make it easier next time to convert to next nginx version.

1.16/Dockerfile
pkubatrh: Will do this in a follow-up PR, where I enable the rest of the image versions (fedora, centos, rhel7)

pkubatrh:
Moved all the other Dockerfiles to 1.16 as well, to avoid confusion. @hhorak @phracek PTAL

1.16/Dockerfile
phracek: This could be documented somewhere in the GitLab repository, I did not know about it.
Like all CentOS repositories which are used by SCL containers.

pkubatrh:
[test-openshift]

1.16/Dockerfile
pkubatrh: I tend to check the postgresql distgen Dockerfile template where we have it hardcoded:
https://github.com/sclorg/postgresql-container/blob/master/src/Dockerfile#L61

pkubatrh:
Added mentions of 1.16 to the root README as well

[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
Note that Fedora tests will fail due to f29 having 1.16 by default now. Will follow up in a separate PR.

pkubatrh:
[test-openshift]

pkubatrh:
There is a hang on Fedora tests, similar to what was encountered in https://github.com/sclorg/container-common-scripts/pull/129 but no idea yet what is the cause of it. Will look into it and fix in the follow-up PR for Fedora enablement (CI would fail anyway).

pkubatrh:
Thanks for review! Merging

PR #101, opened 2020-01-10
Fix openshift.io/display-name version

phracek:
This pull request fixes only versions in openshift.io/display-name.


phracek:
Merging since it is simple change.

Issue #108, opened 2020-04-13
Install mod-http-perl module


chriscarpenter12:
Similar to the [other request for streaming support](https://github.com/sclorg/nginx-container/issues/95) can we install the module to support the perl module?

The default nginx installation is set to load that module dynamically already it just needs the module installed.

```
sh-4.4$ nginx -V
nginx version: nginx/1.16.1
built by gcc 8.3.1 20190507 (Red Hat 8.3.1-4) (GCC) 
built with OpenSSL 1.1.1c FIPS  28 May 2019
TLS SNI support enabled
[...]

PR #109, opened 2020-04-14
Add nginx-mod-http-perl module

chriscarpenter12:
Resolves issue #108 about adding support for the nginx perl module.

### DIFF
--- a/1.12/test/run
+++ b/1.12/test/run
@@ -141,7 +141,7 @@ test_command() {
   if ! echo "${out}" | grep -q "${expected}"; then
     echo "ERROR[exec /bin/sh -ic "${run_cmd}"] Expected '${expected}', got '${out}'"
     return 1
-  fi  
+  fi
 
   return 0
 }
@@ -174,8 +174,7 @@ test_logs() {
   return 0
 }
 
-test_for_output()
-{
+test_for_output() {
   local url="$1"
   local expect="$2"
   local host="${3-localhost}"
@@ -187,9 +186,9 @@ test_for_output()
   info "Testing URL $url with HTTP hostname ${host} produces output $expect"
 
   while [ $attempt -le $max_attempts ]; do
-    response=$(curl -s -H "Host: ${host}" ${url})
-    echo "${response}"
+    response=$(curl -is -H "Host: ${host}" ${url})
     status=$?
+    echo "${response}"
     if [ $status -eq 0 ]; then
       if echo "${response}" | grep -q "${expect}"; then
         info "Success!"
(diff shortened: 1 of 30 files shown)


phracek:
Thanks for the PR! Great approach.

chriscarpenter12:
There's one thing I'm seeing that's missing with this implementation so far. The ability for users to specify their own perl modules because the only want to include files in the s2i build is through the defined config directories or the source directory which you wouldn't want these file to be publicly accessible. Working on a solution for this.

chriscarpenter12:
Finished necessary changes to allow loading user created perl modules from the `perl_modules` and `perl_require` directives. Similar setup to the nginx config directories there's a new `nginx-perl` directory for users to place perl modules.

phracek:
[test]

1.12/test/run
```diff
@@ -187,7 +186,7 @@ test_for_output()
   info "Testing URL $url with HTTP hostname ${host} produces output $expect"
 
   while [ $attempt -le $max_attempts ]; do
-    response=$(curl -s -H "Host: ${host}" ${url})
+    response=$(curl -is -H "Host: ${host}" ${url})
     echo "${response}"
     status=$?
```
phracek, line 191: `$?` is return code of `echo` command.  Is this intended?

1.14/test/run
```diff
@@ -187,7 +186,7 @@ test_for_output()
   info "Testing URL $url with HTTP hostname ${host} produces output $expect"
 
   while [ $attempt -le $max_attempts ]; do
-    response=$(curl -s -H "Host: ${host}" ${url})
+    response=$(curl -is -H "Host: ${host}" ${url})
     echo "${response}"
     status=$?
```
phracek, line 191: The same `$?` is return value of `echo` command. Is it intended?

1.16/test/run
```diff
@@ -187,7 +186,7 @@ test_for_output()
   info "Testing URL $url with HTTP hostname ${host} produces output $expect"
 
   while [ $attempt -le $max_attempts ]; do
-    response=$(curl -s -H "Host: ${host}" ${url})
+    response=$(curl -is -H "Host: ${host}" ${url})
     echo "${response}"
     status=$?
```
phracek, line 191: The same as above

1.12/test/run
chriscarpenter12: Great catch! I didn't even see that. Will updated in a minute.

chriscarpenter12:
> Only minor issues.

Cleaned up those order issues and reran the tests. All passed locally.

phracek:
[test][test-openshift]

chriscarpenter12:
@phracek What is the usual timeframe for merges to hit a release to redhat registry?

notroj:
Thanks a lot @chriscarpenter12 for the contribution!

phracek:
@chriscarpenter12 Approximately every 6 weeks, as far as I know, but It can be a little bit more.

chriscarpenter12:
According to the tags on that container one was released 3 days ago. Do we have to wait 6 weeks then? Was hoping to get these changes :)

PR #111, opened 2020-05-22
Reorder imagestreams

phracek:
Order is now: latest, 1.16, and 1.14.

We would like to test if the latest version is present in the imagestreams.
It is done by ansible-script (sclorg/ansible-tests#4)

The second tag is going to be used for test if the version is the latest.


bparees:
Seems fine to me

PR #114, opened 2020-06-16
Typo having extra dash

saisankargochhayat:
The s2i command in the example fails with three dashes.


phracek:
Thanks for the PR. LGTM.

PR #117, opened 2020-06-25
Run all tests and report results at the end

phracek:
Run all tests and report results at the end.

This commit executes all tests and at the end
reports the whole summary for each version.

### DIFF
--- a/1.14/test/run
+++ b/1.14/test/run
@@ -96,8 +96,10 @@ cleanup() {
 check_result() {
   local result="$1"
   if [[ "$result" != "0" ]]; then
+    cleanup_test_app
     TESTSUITE_RESULT=1
   fi
+  return $result
 }
 
 wait_for_cid() {
@@ -253,12 +255,13 @@ test_application() {
   wait_for_cid
 
   test_scl_usage "nginx -v" "nginx version: nginx/$VERSION."
-  check_result $?
+  check_result $? || return 1
 
   test_connection
-  check_result $?
+  check_result $? || return 1
 
   cleanup_test_app
+  return 0
 }
 
 test_pre_init_script() {
@@ -269,12 +272,13 @@ test_pre_init_script() {
   wait_for_cid
 
   test_command "cat /opt/app-root/etc/nginx.d/default.conf | grep 'resolver' | sed -e 's/resolver\(.*\);/\1/' | grep 'DNS_SERVER'" ""
-  check_result $?
+  check_result $? || return 1
 
   test_connection_s2i
-  check_result $?
+  check_result $? || return 1
 
   cleanup_test_app
+  return 0
 }
 
 test_log_output() {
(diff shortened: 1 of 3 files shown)


phracek:
Report result for RHEL-7 looks like:
```bash

[PASSED] test_s2i_usage
[PASSED] test_docker_run_usage
[PASSED] test_application
[PASSED] test_log_output
[PASSED] test_log_volume_output
[PASSED] test_application_user
[PASSED] test_pre_init_script
[PASSED] test_perl_directive

Tests for rhscl/nginx-114-rhel7 succeeded.

[INFO] All tests finished successfully for rhscl/nginx-114-rhel7....
```
and
```bash
[PASSED] test_s2i_usage
[PASSED] test_docker_run_usage
[PASSED] test_application
[PASSED] test_log_output
[PASSED] test_log_volume_output
[PASSED] test_application_user
[PASSED] test_pre_init_script
[PASSED] test_perl_directive

Tests for rhscl/nginx-116-rhel7 succeeded.

[INFO] All tests finished successfully for rhscl/nginx-116-rhel7....

```

phracek:
[test]

pkubatrh:
Thanks for the PR!

After looking and the diff and seeing a lot of replacements of `check_result` calls by if clauses I wonder, maybe it would be better to re-use the `check_result` function somehow to still have the checks in one place.
Maybe instead of using local variables for results we could just use a global var that would get set by `check_result` on error, and reset on a new test case start?

phracek:
I will rewrite it a bit. Thanks for the Review.

1.14/test/run
```diff
@@ -339,65 +350,68 @@ test_perl_directive() {
...
+  for test_case in $TEST_LIST; do
+    info "Running test $test_case ...."
+    if $test_case; then
+      printf -v test_short_summary "${test_short_summary}[PASSED] $test_case\n"
+    else
+      printf -v test_short_summary "${test_short_summary}[FAILED] $test_case\n"
+      [ -n "${FAIL_QUICKLY:-}" ] && echo "$sum" && return 1
+    fi
+  done;
+  cleanup "${APP_NAME}"
```
pkubatrh, line 378: `APP_NAME` does not seem to be defined anywhere

1.14/test/run
```diff
@@ -339,65 +350,68 @@ test_perl_directive() {
...
+TEST_LIST=${TEST_LIST_PERL_APP} run_all_tests
 
-test_perl_directive
+echo "$test_short_summary"
 
-cleanup "perl-test-app"
+if [ $TESTSUITE_RESULT -eq 0 ] ; then
+  echo "Tests for ${IMAGE_NAME} succeeded."
+else
+  echo "Tests for ${IMAGE_NAME} failed."
```
pkubatrh, line 414: `return $TESTSUITE_RESULT` here?

1.14/test/run
```diff
@@ -339,65 +350,68 @@ test_perl_directive() {
...
+    fi
+  done;
+  cleanup "${APP_NAME}"
+}
 
-# Test application with default uid
-test_application
+function build_image() {
+  local test_name="$1"
+  cid_file=$(mktemp -u --suffix=.cid)
```
pkubatrh, line 383: Not sure if the cid file should be defined once over here, during the build of the image as it is used for runtime containers.

pkubatrh: Though it seems like each test case is doing its cidfile cleanup when necessary, so might be ok.

1.14/test/run
phracek: It has to be exit and not return. Return if from function ;)

1.14/test/run
phracek: Fixed by https://github.com/sclorg/nginx-container/pull/117/commits/4ca4b3b3d45b7b378c289efbf50aa37a8ff5abc2

1.14/test/run
phracek: Fixed by https://github.com/sclorg/nginx-container/pull/117/commits/4ca4b3b3d45b7b378c289efbf50aa37a8ff5abc2

pkubatrh:
Needs to be rebased

pkubatrh:
For the test case functions using `check_result`, can they actually fail now? Looking at the changes I do expect the test suite as a whole failing because of the `TESTSUITE_RESULT` variable being set to 1, but the specific test case will show up as having passed.

phracek:
@pkubatrh Good catch, check_result sets that something failed and at the end the whole test suite failed too, but the alone test case passed.
`check_result` function now return the value 0 / 1 and in relevant testing functions if `check_result` returns none value then the function fails too.

pkubatrh:
lgtm now, thanks!

PR #121, opened 2020-08-21
[DO NOT MERGE] Demonstrate problem in running tests

hhorak:
This demonstrates the same problem as https://github.com/sclorg/postgresql-container/pull/363.


hhorak:
This test suite depends on explicit exit code checking, so the problem demonstrated in this PR is not a real issue actually.

PR #122, opened 2020-08-24
Add CentOS 8 image streams

hhorak:
Let's do not merge yet, there are ongoing discussions whether nginx container image is gonna be released as a UBI -- in that case, we can avoid building CentOS 8 images.

### DIFF
--- a/imagestreams/nginx-centos.json
+++ b/imagestreams/nginx-centos.json
@@ -21,12 +21,32 @@
         },
         "from": {
           "kind": "ImageStreamTag",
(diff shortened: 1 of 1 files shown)


hhorak:
> Let's do not merge yet, there are ongoing discussions whether nginx container image is gonna be released as a UBI -- in that case, we can avoid building CentOS 8 images.

After further discussion it was agreed that it's better to add CentOS 8 imagestream now and once UBI is available, then change it immediately.

PR #132, opened 2020-11-23
add default_server to sed file for rhscl as well

pkubatrh:


1.18/root/opt/app-root/nginxconf.sed
```diff
@@ -1,4 +1,4 @@
-/listen/s%80%8080%
+/listen/s%80%8080 default_server%
```
phracek, line 1: Only one question. This is relevant only for version `1.18`? What about the rest?

phracek (changes requested):

1.18/root/opt/app-root/nginxconf.sed
phracek: It is relevant only for 1.18. Then LGTM.

phracek (approved):
Thanks for the fix. LGTM.

phracek:
[test]

pkubatrh:
The rhel8 failure seems like our known podman issue with CID files, so lest merge. Thanks for review Petr!

PR #139, opened 2021-01-07
Use --chown option for ADD when adding app source instead of changing USER

hhorak:
I found this `--chown` option of `ADD` statement in the Dockerfile today, it seems like much better way to set correct permissions to make the assemble script work.


hhorak:
Hm, it seems to not work with `docker` :(

hhorak:
[test]

hhorak:
Similar PR to https://github.com/sclorg/s2i-python-container/pull/422.

hhorak:
[test]

hhorak:
[test]

pkubatrh:
lgtm, thanks!

PR #141, opened 2021-02-03
Fix calling each testsuite and catch result

phracek:
This commit fixes the catching result of each function.
If the testcase does not exist, then failure is caught.


phracek:
[test]

phracek:
@fila43 Please take a look at it. Thanks.

phracek:
@pkubatrh If you have time, please take a look at it. Thanks.

fila43:
Looks fine, LGTM

PR #142, opened 2021-02-15
Fix link examples dir

phracek:
Add examples directory as symlink into each version.

This will cause the examples directory is also synced into downstream


phracek:
[test]

phracek:
@pkubatrh PTAL. Thanks.

PR #143, opened 2021-03-09
imagestreams: point 1.14 centos image to quay

vrutkovs:
Prefer quay images instead of dockerhub to avoid ratelimiting


phracek:
[test][test-openshift]

phracek:
[test-openshift]

PR #145, opened 2021-04-09
nginx-centos imagestream: fix 1.16 centos8 tag

vrutkovs:
nginx 1.16 CentOS8 tag should refer to CentOS8 image


vrutkovs:
@phracek

PR #156, opened 2021-09-08
Fix GitHub Actions for Fedora and CentOS 7

phracek:


phracek:
This does not affect code. Merging.

PR #154, opened 2021-09-08
Support version 1.20

phracek:

PR #157, opened 2021-09-16
Add examples 1.20

phracek:
This pull request only adds examples for version 1.20.

Missing examples cause testing errors for Fedora https://github.com/sclorg/nginx-container/pull/154
Especially see: https://github.com/sclorg/nginx-container/pull/154#issuecomment-916021333

Why does this happen?

In the test Dockerfile{.s2i} we are cloning https://github.com/nginx-container repository https://github.com/sclorg/nginx-container/blob/master/test/run#L424
and use function  https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L834

But in the case of new versions, the official repository does not contain corresponding version in examples yet ;)

### DIFF
--- a/examples/1.20/test-app/index.html
+++ b/examples/1.20/test-app/index.html
@@ -1,0 +1,8 @@
+<html>
+<head>
+	<title>Test NGINX passed</title>
+</head>
+<body>
+<h1>NGINX is working</h1>
+</body>
+</html>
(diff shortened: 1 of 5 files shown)


phracek:
[test]

fila43:
Tests passed, LGTM

PR #164, opened 2021-12-10
Nginx-1.16 reached EOL in winter 2021

phracek:
As RHEL7 as RHEL8 images reached EOL in winter 2021.

Therefore remove it from the build process.
Also pushing into Quay.io/centos7/nginx-116-centos7 is not done.


phracek:
[test-all]

pkubatrh:
LGTM. The ubi7-based nginx 1.16 image is planned to be EOL at the end of Dec 2021, so we need to wait with merging this PR until January 2022

PR #165, opened 2021-12-10
NginX-1.20 is out and available

phracek:
Also CentOS7 packages are present here:
http://mirror.centos.org/centos/7/sclo/x86_64/rh/Packages/r/

Let's enable builds and also pushing into
quay.io/centos7/nginx-120-centos7

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -38,14 +38,6 @@ jobs:
           ver="${{ matrix.version }}"
           echo "::set-output name=SHORT_VER::${ver//./}"
 
-      - name: Login to Quay.io private registry
-        if: ${{ matrix.registry_namespace != 'centos7' }}
-        uses: redhat-actions/podman-login@v1
-        with:
-          registry: quay.io
-          username: ${{ secrets.QUAY_IMAGE_SCLORG_BUILDER_USERNAME }}
-          password: ${{ secrets.QUAY_IMAGE_SCLORG_BUILDER_TOKEN }}
-
       - name: Check if .exclude-${{ matrix.tag }} is present in version directory
         id: check_exclude_file
         # https://github.com/marketplace/actions/file-existence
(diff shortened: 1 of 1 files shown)


phracek:
[test-all]

phracek:
This request has to be merged after https://github.com/sclorg/nginx-container/pull/164

phracek:
[test-all]

phracek:
[test-all]

phracek:
[test]

phracek:
[test-openshift]

zmiklank:
Would it make sense to split the PR into two separate commits? The change of build-and-push action doesn't seem necessarily related to the Nginx 1.20 release, as the 1.20 has been present in this action even before this PR. Or it wasn't working in the previous form?

phracek:
> Would it make sense to split the PR into two separate commits? The change of build-and-push action doesn't seem necessarily related to the Nginx 1.20 release, as the 1.20 has been present in this action even before this PR. Or it wasn't working in the previous form?

@zmiklank This is partially true. 1.20 is out and we can build-and-push also version 1.20 to quay. But CentOS Stream 9 should not be a part of this PR. I will remove C9S.

phracek:
@zmiklank I see. I have already removed it. But as soon as this pull request is merged, then it will be automatically built and pushed to `quay.io/centos7` organization https://quay.io/repository/centos7/nginx-120-centos7.

Therefore also updating `build-and-push.yml` makes sense, I guess.

phracek:
[test-all]

phracek:
Tests failed because of InfraStructure problems.

Let's try it one more time
[test-all]

phracek:
[test-all]

.github/workflows/build-and-push.yml
```diff
@@ -8,17 +8,17 @@ jobs:
...
+          - version: "1.18"
+            dockerfile: "Dockerfile"
+            registry_namespace: "centos7"
+            tag: "centos7"
+          - version: "1.20"
+            dockerfile: "Dockerfile"
             registry_namespace: "centos7"
-            registry_suffix: "-centos7"
             tag: "centos7"
-          - dockerfile: "Dockerfile.c9s"
```
zmiklank, line 18: Why the c9s option is removed in this commit? Does it have an impact on building and pushing Nginx 1.20 container?

.github/workflows/build-and-push.yml
```diff
@@ -38,14 +38,6 @@ jobs:
           ver="${{ matrix.version }}"
           echo "::set-output name=SHORT_VER::${ver//./}"
 
-      - name: Login to Quay.io private registry
```
zmiklank, line 41: Same question as for line 18: does removing this have some effect on building and pushing the Nginx 1.20 container?

phracek: The same as above

zmiklank (approved):

PR #170, opened 2022-02-15
Do not end line from readlink with NUL

zmiklank:
Source: https://github.com/sclorg/s2i-perl-container/pull/216


phracek:
This affects only container tests.

[test]

PR #173, opened 2022-02-28
use new build-and-push action

zmiklank:
Using the new build-and-push-action [1].

[1] https://github.com/sclorg/build-and-push-action

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -1,99 +1,42 @@
-name: Build and push CentOS7 images to Quay.io registry
+name: Build and push container images to quay.io registry
 on:
   push:
     branches:
       - master
-
 jobs:
   build-and-push:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
         include:
-          - version: "1.18"
+          - dockerfile_path: "1.18"
             dockerfile: "Dockerfile"
             registry_namespace: "centos7"
+            quayio_username: "QUAY_IMAGE_BUILDER_USERNAME"
+            quayio_token: "QUAY_IMAGE_BUILDER_TOKEN"
             tag: "centos7"
-          - version: "1.20"
+          - dockerfile_path: "1.20"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push.yml
```diff
@@ -1,99 +1,42 @@
...
-      - name: Print image url
-        if: ${{ matrix.registry_namespace == 'centos7' }}
-        run: echo "Image pushed to ${{ steps.push-to-quay-centos7.outputs.registry-paths }}"
-
-      - name: Print image url
-        if: ${{ matrix.registry_namespace != 'centos7' }}
-        run: echo "Image pushed to ${{ steps.push-to-quay-sclorg.outputs.registry-paths }}"
+          registry: "quay.io"
+          registry_namespace: ${{ matrix.registry_namespace }}
+          registry_username: ${{ secrets[matrix.quayio_username] }}
```
phracek, line 38: I would prefer to not add secrets into the matrix but split it into two actions.
One for centos7 and one for c9s based on the matrix.tag.

But this is an idea. WDYT? Does it make sense to you?

.github/workflows/build-and-push.yml
```diff
@@ -1,99 +1,42 @@
...
+          registry_token: ${{ secrets[matrix.quayio_token] }}
```
phracek, line 39: I thinks the format should be:
${{ secrets.[matrix.quayio_token] }} right?
I guess, '.' is missing.

phracek (changes requested):

.github/workflows/build-and-push.yml
phracek: But this is not needed. Based on the later comment. Sorry.

.github/workflows/build-and-push.yml
zmiklank: I was inspired by [this article](https://sbulav.github.io/terraform/github-actions-matrix-secrets/).
The dot is not used there.

phracek: I see. ok.

PR #177, opened 2022-03-08
get PR number from github's event envvar

zmiklank:
Using GitHub's embedded event number variable would save us one step in the GitHub action.
Source:
https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pullrequestevent


phracek (approved):
LGTM

PR #181, opened 2022-03-16
update container-tests to use newest testing farm action

zmiklank:
This can be merged only when TFGA v1.2.9 is released (with this change integrated: https://github.com/sclorg/testing-farm-as-github-action/pull/40)

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -17,22 +17,39 @@ jobs:
             os_test: "fedora"
             context: "Fedora"
             compose: "CentOS-7"
+            api_key: "TF_PUBLIC_API_KEY"
+            branch: "main"
+            tmt_repo: "https://github.com/sclorg/sclorg-testing-farm"
           - tmt_plan: "centos7"
             os_test: "centos7"
             context: "CentOS7"
             compose: "CentOS-7"
+            api_key: "TF_PUBLIC_API_KEY"
+            branch: "main"
+            tmt_repo: "https://github.com/sclorg/sclorg-testing-farm"
           - tmt_plan: "rhel7-docker"
             os_test: "rhel7"
             context: "RHEL7"
             compose: "RHEL-7.9-Released"
+            api_key: "TF_INTERNAL_API_KEY"
+            branch: "master"
+            tmt_repo: "https://gitlab.cee.redhat.com/platform-eng-core-services/sclorg-tmt-plans"
+            tf_scope: "private"
           - tmt_plan: "rhel8-docker"
             os_test: "rhel8"
             context: "RHEL8"
             compose: "RHEL-8.3.1-Released"
+            api_key: "TF_INTERNAL_API_KEY"
+            branch: "master"
+            tmt_repo: "https://gitlab.cee.redhat.com/platform-eng-core-services/sclorg-tmt-plans"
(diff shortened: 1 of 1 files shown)


zmiklank:
The TF-as-GithubAction is in the required state, so this can be merged (after a review).

.github/workflows/container-tests.yml
```diff
@@ -17,22 +17,42 @@ jobs:
             os_test: "fedora"
             context: "Fedora"
             compose: "CentOS-7"
+            api_key: "TF_PUBLIC_API_KEY"
+            branch: "main"
+            tmt_repo: "https://github.com/sclorg/sclorg-testing-farm"
+            tf_scope: "public"
```
phracek, line 23: This is not needed, I guess. The default value is public.

.github/workflows/container-tests.yml
```diff
@@ -44,33 +64,16 @@ jobs:
...
-          echo "::set-output name=API_KEY::$api_key"
-          echo "::set-output name=BRANCH_NAME::$branch_name"
-          echo "::set-output name=TMT_REPO::$tmt_repo"
-        shell: bash
-
       # https://github.com/sclorg/testing-farm-as-github-action
       - name: Schedule tests on external Testing Farm by Testing-Farm-as-github-action
         id: github_action
-        uses: sclorg/testing-farm-as-github-action@v1
+        uses: sclorg/testing-farm-as-github-action@v1.2.9
```
phracek, line 70:
```suggestion
        uses: sclorg/testing-farm-as-github-action@v1
```

phracek (changes requested):
I guess, tf_scope: 'public' is not needed. It is set by default.

.github/workflows/container-tests.yml
zmiklank: Yes, but we take `tf_scope` input from the matrix. And in the matrix, there is no default. So IMHO empty value would be passed to the action, or it will fail.

phracek: Undefined `tf_scope` in the matrix works properly. See https://github.com/sclorg/s2i-base-container/pull/209.
And the reason is simple. Only if `tf_scope` is defined as `private` then `osci...` is used otherwise `public` URL.

phracek (approved):
LGTM. As soon as tests are passing, you can merge it.

.github/workflows/container-tests.yml
```diff
@@ -44,33 +64,16 @@ jobs:
...
         id: github_action
         uses: sclorg/testing-farm-as-github-action@v1
         with:
-          api_key: ${{ steps.tf_values.outputs.API_KEY }}
-          git_url: ${{ steps.tf_values.outputs.TMT_REPO }}
-          git_ref: ${{ steps.tf_values.outputs.BRANCH_NAME }}
+          api_key: ${{ secrets[matrix.api_key] }}
+          git_url: ${{ matrix.tmt_repo }}
+          git_ref: ${{ matrix.branch }}
+          tf_scope: ${{ matrix.tf_scope }}
```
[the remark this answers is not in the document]
zmiklank: Sorry, I do not understand, why the `tf_scope` should be removed also from actions inputs. Wouldn't then always be used the Public Testing Farm?

Issue #166, opened 2022-01-07
alarming comment in Dockerfile questions safety of chmod use


mhrivnak:
A [comment in the Dockerfile](https://github.com/sclorg/nginx-container/blob/c7dda5b/1.20/Dockerfile.rhel8#L70) says ` # FIXME: Not sure if this is safe to do, just a hack to make the image work`. It then runs `chmod -R a+rwx /var/run`. That is all visible on the [Red Hat Container Catalog](https://catalog.redhat.com/software/containers/ubi8/nginx-120/6156abfac739c0a4123a86fd?container-tabs=dockerfile)'s Dockerfile display, where it is not confidence-inspiring.

zmiklank:
Thanks for creating the issue.
Should be fixed in https://github.com/sclorg/nginx-container/pull/205.

PR #185, opened 2022-04-11
Support C8S for nginx-container

phracek:
This pull request adds support CentOS Stream 8 for nginx-container.
It also updates README.md files and build-and-push action.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -19,22 +19,39 @@ jobs:
             quayio_username: "QUAY_IMAGE_BUILDER_USERNAME"
             quayio_token: "QUAY_IMAGE_BUILDER_TOKEN"
             tag: "centos7"
+            suffix: "centos7"
           - dockerfile_path: "1.20"
             dockerfile: "Dockerfile"
             registry_namespace: "centos7"
             quayio_username: "QUAY_IMAGE_BUILDER_USERNAME"
             quayio_token: "QUAY_IMAGE_BUILDER_TOKEN"
             tag: "centos7"
+            suffix: "centos7"
           - dockerfile_path: "1.20"
             dockerfile: "Dockerfile.c9s"
             registry_namespace: "sclorg"
             quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
             quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
             tag: "c9s"
+            suffix: "c9s"
+          - dockerfile_path: "1.20"
+            dockerfile: "Dockerfile.c8s"
+            registry_namespace: "sclorg"
+            quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
+            quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
+            tag: "c8s"
+            suffix: "c8s"
+          - dockerfile_path: "1.18"
+            dockerfile: "Dockerfile.c8s"
+            registry_namespace: "sclorg"
+            quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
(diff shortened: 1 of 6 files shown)


phracek:
[test]

phracek:
[test-openshift]

.github/workflows/build-and-push.yml
```diff
@@ -44,3 +61,4 @@ jobs:
           docker_context: ${{ matrix.dockerfile_path }}
           dockerfile_path: ${{ matrix.dockerfile_path }}
           tag: ${{ matrix.tag }}
+          suffix: ${{ matrix.suffix }}
```
zmiklank, line 64:
```suggestion
          suffix: ${{ matrix.tag }}
```

.github/workflows/build-and-push.yml
```diff
@@ -19,22 +19,39 @@ jobs:
...
             quayio_token: "QUAY_IMAGE_BUILDER_TOKEN"
             tag: "centos7"
+            suffix: "centos7"
           - dockerfile_path: "1.20"
             dockerfile: "Dockerfile.c9s"
             registry_namespace: "sclorg"
             quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
             quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
             tag: "c9s"
+            suffix: "c8s"
```
zmiklank, line 36:
```suggestion
            suffix: "c9s"
```
But I would rather won't state suffix here at all, as it seems, that always stays that suffix=tag.

1.20/Dockerfile.c8s
```diff
@@ -0,0 +1,81 @@
...
+    mkdir -p ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
+    mkdir -p ${NGINX_LOG_PATH} && \
+    mkdir -p ${NGINX_PERL_MODULE_PATH} && \
+    chmod -R a+rwx ${NGINX_APP_ROOT}/etc && \
+    chmod -R a+rwx /var/lib/nginx && \
+    chmod -R a+rwx ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
+    chown -R 1001:0 ${NGINX_APP_ROOT} && \
+    chown -R 1001:0 /var/lib/nginx && \
+    chown -R 1001:0 ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
+    # FIXME: Not sure if this is safe to do, just a hack to make the image work
```
zmiklank, line 69: Shouldn't we address this further?

zmiklank (changes requested):

.github/workflows/build-and-push.yml
phracek: Please not, when we will support Fedora build-and-push, the `tag` will not be used.

phracek: Or better say, the suffix will not be used at all.

1.20/Dockerfile.c8s
phracek: Good question. I don't not know who add this sentence, but it does not work without it.

zmiklank: An issue exists for this: https://github.com/sclorg/nginx-container/issues/166

phracek: Addressed here: https://github.com/sclorg/nginx-container/pull/187

phracek: This should be addressed by the other pull request and not by this one.

zmiklank: Agreed.

zmiklank (commented):
Please rename this commit: [Update .github/workflows/build-and-push.yml](https://github.com/sclorg/nginx-container/pull/185/commits/c4498dbf4f67d4c4a7336c2de5dcf4e0b39ea628) or sqash it with some other before merging.

Else LGTM.

phracek:
[test-all]

PR #187, opened 2022-04-13
remove comment

phracek:
This pull request removes comment which is not needed

Closes: #166 

### DIFF
--- a/1.18/Dockerfile.rhel8
+++ b/1.18/Dockerfile.rhel8
@@ -67,7 +67,6 @@ RUN sed -i -f ${NGINX_APP_ROOT}/nginxconf-fed.sed ${NGINX_CONF_PATH} && \
     chown -R 1001:0 ${NGINX_APP_ROOT} && \
     chown -R 1001:0 /var/lib/nginx && \
     chown -R 1001:0 ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
-    # FIXME: Not sure if this is safe to do, just a hack to make the image work
     chmod -R a+rwx /var/run && \
     chown -R 1001:0 /var/run && \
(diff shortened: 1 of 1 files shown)


phracek:
[test-all]

pkubatrh:
This PR is missing any comment on whether the command is actually ok to use in the Dockerfile or not. We should not just blindly remove comments about possible impacts on security without checking the impact.

zmiklank:
`/var/run` is symlink to `/run`: 
```
bash-5.1$ ls -al /var/run
lrwxrwxrwx. 1 default root 6 Apr 11 10:39 /var/run -> ../run
```
These two lines 
```
   chmod -R a+rwx /var/run && \
   chown -R 1001:0 /var/run && \
```
cause following behaviour:
1) the symlink `/var/run` has owner:group 1001:0, the directory `/run` with all its contents remain in a ownership of root:root.
2) all files existing in the time when this command is executed have 777 permissions. Permission of symlinks in Linux systems can not be changed.

This seems really inconsistent to me and I could not find a reason or use-case in which this could be correct.

Moreover the warning comment is in following Dockerfiles:
```
1.14/Dockerfile.rhel8
1.16/Dockerfile.rhel8
1.18/Dockerfile.rhel8
1.18/Dockerfile.c8s
1.20/Dockerfile.rhel8
1.20/Dockerfile.c8s
1.20/Dockerfile.c9s
```
and this PR addresses only one of them.
Other container images seem to run even without such a significant change of rights.

For example `1.20/Dockerfile.fedora` contains command `chmod -R a+rwx /var /run`, which is also strong, but not that confusing.

I suggest to investigate this further and ideally at least unite the Dockerfiles, wherever possible.

PR #190, opened 2022-04-25
Change permissions in Fedora image

zmiklank:
This is a proposal for fixing https://github.com/sclorg/nginx-container/issues/166 and closing https://github.com/sclorg/nginx-container/pull/187.
For now it is only for v1.20 in Fedora, but can probably be extended also to other distributions, if it passes a code-review.

Explanation can be found in commit messages.

### DIFF
--- a/1.20/Dockerfile.fedora
+++ b/1.20/Dockerfile.fedora
@@ -55,25 +55,35 @@ COPY ./s2i/bin/ $STI_SCRIPTS_PATH
 # Copy extra files to the image.
 COPY ./root/ /
 
-# In order to drop the root user, we have to make some directories world
-# writable as OpenShift default security model is to run the container under
-# random UID.
+# Changing ownership and user rights to support following use-cases:
+# 1) running container on OpenShift, whose default security model
+#    is to run the container under random UID, but GID=0
+# 2) for working root-less container with UID=1001, which does not have
+#    to have GID=0
+# 3) for default use-case, that is running container directly on operating system,
+#    with default UID and GID (1001:1)
+# Supported combinations of UID:GID are thus following:
+# UID=1001 && GID=0
+# UID=<any>&& GID=0
+# UID=1001 && GID=<any>
 RUN sed -i -f ${NGINX_APP_ROOT}/nginxconf-fed.sed ${NGINX_CONF_PATH} && \
-    chmod a+rwx ${NGINX_CONF_PATH} && \
     mkdir -p ${NGINX_APP_ROOT}/etc/nginx.d/ && \
     mkdir -p ${NGINX_APP_ROOT}/etc/nginx.default.d/ && \
     mkdir -p ${NGINX_APP_ROOT}/src/nginx-start/ && \
     mkdir -p ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
     mkdir -p ${NGINX_LOG_PATH} && \
     mkdir -p ${NGINX_PERL_MODULE_PATH} && \
-    chmod -R a+rwx ${NGINX_APP_ROOT}/etc && \
-    chmod -R a+rwx /var /run && \
-    chmod -R a+rwx ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
-    chown -R 1001:0 ${NGINX_APP_ROOT} && \
-    chown -R 1001:0 /var && \
+    chown -R 1001:0 ${NGINX_CONF_PATH} && \
+    chown -R 1001:0 ${NGINX_APP_ROOT}/etc && \
+    chown -R 1001:0 ${NGINX_APP_ROOT}/src/nginx-start/  && \
     chown -R 1001:0 ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
+    chown -R 1001:0 /var/lib/nginx /var/log/nginx /run && \
+    chmod    ug+rw  ${NGINX_CONF_PATH} && \
+    chmod -R ug+rwX ${NGINX_APP_ROOT}/etc && \
+    chmod -R ug+rwX ${NGINX_APP_ROOT}/src/nginx-start/  && \
+    chmod -R ug+rwX ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
+    chmod -R ug+rwX /var/lib/nginx /var/log/nginx /run && \
     rpm-file-permissions
-
 
 USER 1001
 


zmiklank:
[test]

1.20/s2i/bin/assemble
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -e
```
phracek, line 3: This has to be set. In case the `assemble` script is failing for some reason, the build should not continue. Even if `cp -Rf /tmp/src/....` is failing.

zmiklank: Makes sense. I was also thinking, that with unsetting exit-on-error only here, I can cause confusion, because it is present everywhere else.
I did this, because having to remove `set -x` option from multiple files in order to find out, why the tests are failing makes the debugging situation a bit difficult.
In any way, you are true, that I should remove this change, thx.

1.20/Dockerfile.fedora
```diff
@@ -59,19 +59,17 @@ COPY ./root/ /
...
 # random UID.
 RUN sed -i -f ${NGINX_APP_ROOT}/nginxconf-fed.sed ${NGINX_CONF_PATH} && \
-    chmod a+rwx ${NGINX_CONF_PATH} && \
     mkdir -p ${NGINX_APP_ROOT}/etc/nginx.d/ && \
     mkdir -p ${NGINX_APP_ROOT}/etc/nginx.default.d/ && \
     mkdir -p ${NGINX_APP_ROOT}/src/nginx-start/ && \
     mkdir -p ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
     mkdir -p ${NGINX_LOG_PATH} && \
     mkdir -p ${NGINX_PERL_MODULE_PATH} && \
+    chmod a+rwx ${NGINX_CONF_PATH} && \
```
phracek, line 68: Why it was moved above?

phracek: Only from reason to have chmod togther?

1.20/Dockerfile.fedora
```diff
@@ -59,19 +59,17 @@ COPY ./root/ /
...
     chmod -R a+rwx ${NGINX_APP_ROOT}/etc && \
-    chmod -R a+rwx /var /run && \
+    chmod -R a+rwx ${NGINX_APP_ROOT}/src/nginx-start/  && \
```
phracek, line 70: What about to have 
```suggestion
    chmod -R a+rwx ${NGINX_APP_ROOT}/src && \
```

1.20/Dockerfile.fedora
```diff
@@ -59,19 +59,17 @@ COPY ./root/ /
...
     chmod -R a+rwx ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
-    chown -R 1001:0 ${NGINX_APP_ROOT} && \
```
phracek, line 72: This is needed because of at the end we are switching to USER 1000
https://github.com/sclorg/nginx-container/blob/master/1.20/Dockerfile.fedora#L78

1.20/Dockerfile.fedora
```diff
@@ -59,19 +59,17 @@ COPY ./root/ /
...
-    chown -R 1001:0 /var && \
-    chown -R 1001:0 ${NGINX_CONTAINER_SCRIPTS_PATH}/nginx-start && \
```
phracek, line 74: Are you sure about removing this line? 
This line is executed by this code https://github.com/sclorg/nginx-container/blob/master/1.20/s2i/bin/run#L9, which means, that `USER 1001` runs it.

phracek (changes requested):

1.20/Dockerfile.fedora
zmiklank: yes

1.20/Dockerfile.fedora
zmiklank: It is not.

1) Our tests run with user 10001, not 1001 and they work. However, I do not know, if that was intended: https://github.com/sclorg/nginx-container/blob/59508a8dc90458916926e5ff63762e8ab651cbc4/test/run#L72

2) the `${NGINX_APP_ROOT}` only has src/nginx-start and etc/ directories, and they have been already chmoded to 777.

3) As written in commit message, chown is IMO useless for use in openshift.

1.20/Dockerfile.fedora
zmiklank: That should also be OK. But I wanted to go with minimal changes to the permissions.

1.20/Dockerfile.fedora
phracek: Yeah, it was intended. By default, OpenShift runs a container as an arbitrary user. Like in our case. The number is or can be 100004557 or something like that.

1.20/Dockerfile.fedora
zmiklank: This is the same as issue you mentioned in the comment above.

1.20/Dockerfile.fedora
pkubatrh: Now I see that we also chmod 777 the directories inside. Maybe we should revisit if it is really necessary to give permissions to everyone?

zmiklank: I suppose, this is because we do not know, under which user the container is to be run in openshift environment.

pkubatrh: We know that the user is part of the root group, hence why the `chown -R 1001:0`. But I think the chmod 777 commands were added historically before we had that guarantee? Maybe @hhorak would remember

zmiklank: Hmm, isn't GID also assigned randomly? [1]

[1] https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids

pkubatrh: From the document you linked: "The Container user is always a member of the root group, so it can read or write files accessible by GID=0."
So I do not think anything changed there.

zmiklank: Oh, I see, thanks for noticing. 
That changes the situation. I need to think about the proceeding then.

zmiklank: I have proposed new solution.

1.20/s2i/bin/assemble
```diff
@@ -20,7 +20,6 @@ if [ -d ./nginx-cfg ]; then
     cp -av ./nginx-cfg/*.conf "${NGINX_CONFIGURATION_PATH}"
     rm -rf ./nginx-cfg
   fi
-  chmod -Rf g+rw ${NGINX_CONFIGURATION_PATH}
```
pkubatrh, line 23: Are we sure we can remove this? This is likely in the assemble script since we know the new config files added through s2i will have wrong permissions.

zmiklank:
[test-all]

1.20/s2i/bin/assemble
zmiklank: It works because the assembly script in our tests is run by the 1001:0 user, and thus the files copied should be owned by 1001:0 by default. Using `cp -a` preserves permissions on copied files, which already have rw for group. 

So IMO these chmods in assembly scripts do not have any meaning (please correct me, if I understand this incorrectly) and can confuse users.

pkubatrh: I do not see much need for it either, but I might be missing something. I see the chmod calls were added by @hhorak in #41 so maybe he would remember the reason for adding it?
Considering that in the best case the calls are functionally-wise a no-op (since we expect all the permissions are already set), we might as well keep it in in as it would not make any difference.

hhorak: I'll probably not bring too much light on this -- my #41 was built on top of #32, so I mostly took over what was prepared by @kosciCZ back then. I'm wondering how much legal it is to run the assemble script by root -- I can imagine a user wants not only run the assemble but also install some RPMs before -- in that case few more chown calls might do the difference. (it's just and idea so far, but as @pkubatrh mentioned, maybe ok to keep it there if it shouldn't do anything in most common cases)

zmiklank: I would say, that the assembly script can be run as root, when using rootfull container, with unspecified UID - then the default is AFAIK 0 (root).
However, it would not have to be a problem as the copied configuration files and directories have their permissions changed before, with `fix-permissions` here: https://github.com/sclorg/nginx-container/blob/a4db1d75aaac92c590ecebdc10b8c52d869444c3/1.20/s2i/bin/assemble#L9
That basically does g+rw on $1 (and that is what I have removed from the assembly script).

Anyway, I understand your point of view and if you think, that it would be safer to let those commands in assembly script I would agree, as I can not see how they could be harmful in any way.

zmiklank: I have removed the changes in the assembly script.

zmiklank:
I have addressed all discussed issues. Please re-review, if the tests pass:

[test-all]

zmiklank:
The tests have failed because of infra problems, let's retest:
[test-all]

phracek:
[test-all]

PR #192, opened 2022-04-27
[WIP] Make tests work also for rootless containers

zmiklank:
- add -P to docker run when running container tests

Publish all exposed ports to random ports on the host interface.
Add function for getting actual mapped port, and do not use hardcoded
8080. Return 8080, only when port is not published (works in rootfull
configuration in one subnet).

- return "localhost" as an IP, if container itself does not have any IP
  assigned (when running in rootless state)

- known issues -> rootfull configuration seem to run only for localhost:published_port combination
- also needs support from container-common-scripts (WIP)

NOT INTENDED FOR MERGING YET.

### DIFF
--- a/test/run
+++ b/test/run
@@ -43,8 +43,9 @@ container_exists() {
 }
 
 container_ip() {
-  local ip=$(docker inspect --format="{{ .NetworkSettings.IPAddress }}" $(cat $cid_file))
-  echo ${ip:-localhost}
+  # local ip=$(docker inspect --format="{{ .NetworkSettings.IPAddress }}" $(cat $cid_file))
+  # echo ${ip:-localhost}
+  echo localhost
 }
 
 test_port() {
--- a/test/run
+++ b/test/run
@@ -29,8 +29,6 @@ TEST_LIST_PERL_APP="test_perl_directive"
 
 . $test_dir/test-lib.sh
 
-# TODO: This should be part of the image metadata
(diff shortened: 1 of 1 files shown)


zmiklank:
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

test/run
```diff
@@ -45,7 +43,14 @@ container_exists() {
 }
 
 container_ip() {
-  docker inspect --format="{{ .NetworkSettings.IPAddress }}" $(cat $cid_file)
+  # local ip=$(docker inspect --format="{{ .NetworkSettings.IPAddress }}" $(cat $cid_file))
+  # echo ${ip:-localhost}
+  echo localhost
+}
+
+test_port() {
```
phracek, line 51: This function could be a part of container-common-scripts.
The first parameter would be port and the second parameter would be a protocol

The results would be `echo` as you already have it.

zmiklank: Sounds good. I will need to make time to get back to this soon and get this PR to a mergeable state.

phracek:
[test-all]

phracek:
[test-openshift]

zmiklank:
[test-openshift]

zmiklank:
[test-openshift]

zmiklank:
[test-openshift]

Issue #96, opened 2019-10-10
Containers cannot be restarted


Richard-Degenne:
Hello,

I have encountered an issue which makes containers unable to restart once stopped.

This bug affects versions `1.10`, `1.12` and `1.14`.

# Steps to reproduce

- Build a minimal image.

```
mkdir s2i-nginx-bug
cd s2i-nginx-bug
s2i build . centos/nginx-112-centos7 s2i-nginx-bug:dev
```

- Create a container using the built image.

```
docker run -d --name s2i-nginx-bug s2i-nginx-bug:dev
```

- Observe that the container is up and running.

```
docker ps -a -f name=s2i-nginx-bug

CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                NAMES
72ebd4e6bc9e        s2i-nginx-bug:dev   "container-entrypoin…"   6 seconds ago       Up 5 seconds        8080/tcp, 8443/tcp   s2i-nginx-bug
```

- Restart the container.

```
docker restart s2i-nginx-bug
```

- Observe that the container crashed.

```
docker ps -a -f name=s2i-nginx-bug

CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS                     PORTS               NAMES
72ebd4e6bc9e        s2i-nginx-bug:dev   "container-entrypoin…"   49 seconds ago      Exited (1) 4 seconds ago                       s2i-nginx-bug
```

# Notes

- Inspect the container's logs.

```
docker logs s2i-nginx-bug

/bin/ln: failed to create symbolic link '/var/opt/rh/rh-nginx112/log/nginx/access.log': File exists
```

The error is caused by [the logging feature in the `run` script][run]. The first time the container starts, the `ln` commands run just fine. However, when the container restarts, the targets already exist, resulting in the error shown above.

[run]: https://github.com/sclorg/nginx-container/blob/master/1.12/s2i/bin/run#L11-L14

## Suggested fix

Add the `-f` option to `ln` commands in order to force the recreation of the links.

```sh
if [ ! -v NGINX_LOG_TO_VOLUME -a -v NGINX_LOG_PATH ]; then
    /bin/ln -sf /dev/stdout ${NGINX_LOG_PATH}/access.log
    /bin/ln -sf /dev/stderr ${NGINX_LOG_PATH}/error.log
fi
```

Alternatively, you can check whether the targets already exist beforehand and skip the commands if they do.

## Workaround

[...]

PR #193, opened 2022-05-05
Fix issue #96

Richard-Degenne:


jemerald:
Could you update other versions too?

Richard-Degenne:
I've updated versions `1.10` through `1.20`. I took a look at versions `1.6` and `1.8` but I don't think the issue is applicable there.

zmiklank:
[test-all]

phracek (approved):
LGTM, Although only 1.18 and 1.20 are supported.

PR #195, opened 2022-05-11
Fixes #194: Do not build CentOS Stream 8 into quay.io.

### LINKED ISSUE #194: CentOS Stream 8 should be deprecated
CentOS Stream 8 should be deprecated because we have UBI image.

phracek:
Do not build CentOS Stream 8 into quay.io.

Container tests and OpenShift tests are still present


phracek:
[test-all]

phracek:
[test-all]

phracek:
@pkubatrh Question is if we want to also test the image for CentOS Stream 8 or remove CentOS Stream 8 completely.

pkubatrh:
I do not think we need to test it if we do not provide it at all.

phracek:
I have removed CentOS Stream 8 completely. Including Dockerfile.c8s

phracek:
[test-all]

phracek:
[test-all]

PR #198, opened 2022-06-07
remove eol versions of nginx

zmiklank:
I got to a task where all Dockerfiles in all supported versions need to be adjusted. Then I noticed, that there are many unsupported versions of Nginx in this repository.

The policy about removing old image files is AFAIK not strict so the deprecated image files are to be removed after some time passed since their deprecation.

Has the time come for removing the following versions, in order to clean the repository a little?
If not, I will close the PR, because I am not sure. So please take the PR only as a proposition. 

--

remove all version, for which containers images are no longer built
removed versions:
  1.6/
  1.8/
  1.10/
  1.12/
  1.14/
  1.16/


pkubatrh:
[test]

phracek:
Fedora failure is fixed by https://github.com/sclorg/nginx-container/pull/201

phracek:
[test-openshift]

PR #199, opened 2022-06-08
Exit oc tests if login fails

zmiklank:
The nounset shell option needs to be set after the OC login attempt,
because (from man shopt):
```
-u
  If expansion is attempted on an unset  variable  or
  parameter,  the  shell prints  an  error  message, and, if not
  interactive exits with a non-zero status.
```
That is in this matter inconvenient, as we do not really know from
what has happened if the oc login fails and moreover the test suite
with success, as only the first failed test sets TESTSUITE_RESULT to 1.
If we check login with the ct_os_check_login function, it handles the
situation conveniently.

### DIFF
--- a/test/run-openshift-remote-cluster
+++ b/test/run-openshift-remote-cluster
@@ -18,15 +18,15 @@ test_nginx_integration
 test_nginx_imagestream
 "
 
-set -u
+trap ct_os_cleanup EXIT SIGINT
 
 ct_os_set_ocp4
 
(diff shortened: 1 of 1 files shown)


zmiklank:
[test-all]

phracek (approved):
LGTM. But lets merge it after https://github.com/sclorg/nginx-container/pull/200

phracek:
Let's merge it. OpenShift 4 issue is solved by https://github.com/sclorg/nginx-container/pull/200

PR #204, opened 2022-06-14
Fix symlink for the test app

hhorak:
This is just a minor fix for a copy&paste error. No real implications, as the example apps are the same.


hhorak:
[test]

hhorak:
Simple fix that likely does not need review.

PR #221, opened 2022-10-21
Use the same context for Build and Push action as it is used in the Makefile

hhorak:
Not sure how to test this before merging though. @zmiklank, please, take a look.


hhorak:
[test]

zmiklank (approved):

PR #222, opened 2022-10-31
Use nss_wrapper-libs only to get rid of perl

hhorak:
This is not possible in RHEL8 yet unfortunately, as there is no `nss_wrapper-libs` package available.


hhorak:
[test]

phracek:
[test-openshift]

phracek:
> This is not possible in RHEL8 yet unfortunately, as there is no `nss_wrapper-libs` package available.

@hhorak I don't follow, `nss_wrapper-libs` is not available in RHEL8 but you modified Dockerfile.fedora?
Is it intended?

phracek (approved):
LGTM

PR #223, opened 2022-11-10
tests: proper removal of containers after tests

zmiklank:
* the cid_file needs to be created so that ct_cleanup function can clean up correctly
* the container, that does not use cid_file has to be run with --rm parameter, so it cleans after itself

### DIFF
--- a/test/run
+++ b/test/run
@@ -121,7 +121,7 @@ test_s2i_usage() {
 
 test_docker_run_usage() {
   info "Testing 'docker run' usage"
-  docker run ${IMAGE_NAME} &>/dev/null
+  docker run --rm ${IMAGE_NAME} &>/dev/null
 }
 
 test_scl_usage() {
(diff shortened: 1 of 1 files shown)


zmiklank:
[test-all]

test/run
```diff
@@ -427,6 +427,7 @@ TEST_SUMMARY=''
 ct_enable_cleanup
 
 CID_FILE_DIR=$(mktemp -d)
+cid_file="$CID_FILE_DIR/$(mktemp -u -p . --suffix .cid)"
```
phracek, line 430: I don't follow, why this? Can you please paste a link to why cid_file should be created?

phracek (changes requested):

phracek:
Let's test CentOS7 with new plans once again.

[test-all]

test/run
zmiklank: In the `test/run` is the `cid_file` always used for storing the id of created container. Untill now this variable has not been set properly.
Example: https://github.com/sclorg/nginx-container/blob/ac850b9244dff752a472cee429aab88e3ea30fd0/test/run#L80

phracek (approved):

PR #224, opened 2022-11-10
use new init function from common/ to start test environment

zmiklank:


zmiklank:
Depends on https://github.com/sclorg/container-common-scripts/pull/306

zmiklank:
[test-all]

phracek (approved):
I did not hit any issues.

PR #227, opened 2022-11-15
Add a test in Dockerfile directly to check the version

hhorak:
Otherwise when a wrong stream is installed, it's only found by scl test, which is too late in the process and might even go through uncought, especially when the scl test is removed for non-SCL distros at some point.


hhorak:
[test]

phracek:
@hhorak CentOS 7 test has failed:
```bash
9dca3059a166df29b1762d3d0bca3244c5708a69a7b108a77278ff3bdd7909c3
STEP 9: COPY $NGINX_VERSION/s2i/bin/ $STI_SCRIPTS_PATH
Error: error dry-running "COPY $NGINX_VERSION/s2i/bin/ $STI_SCRIPTS_PATH": no files found matching "/root/sclorg/nginx-container/s2i/bin": no such file or directory
make[1]: *** [1.20] Error 125
```

It looks like we have to use HARDCODED version

phracek:
rebased

[test-all]

phracek (approved):
Thanks for the PR! LGTM. Tests are passing.

PR #228, opened 2022-11-16
centos7: use hardcoded VERSION in COPY command in Dockerfile

zmiklank:
If we leave $NGINX_VERSION there the command fails, because $NGINX_VERSION is substituted with "".
This is caused by docker->podman change.
This fix is only workaround.
Strange is, that when I print value of $NGINX_VERSION directly in the Dockerfile it is set correctly to either 1.20 or 1.18.
    
The reason of this issue in not a invalid build context, the behavior would be different then.
Unfortunately it seems that all other sclorg containers are affected by this too.

### DIFF
--- a/1.18/Dockerfile
+++ b/1.18/Dockerfile
@@ -59,10 +59,10 @@ RUN yum install -y yum-utils gettext hostname && \
     yum -y clean all --enablerepo='*'
 
 # Copy the S2I scripts from the specific language image to $STI_SCRIPTS_PATH
-COPY $NGINX_VERSION/s2i/bin/ $STI_SCRIPTS_PATH
+COPY 1.18/s2i/bin/ $STI_SCRIPTS_PATH
 
 # Copy extra files to the image.
-COPY $NGINX_VERSION/root/ /
+COPY 1.18/root/ /
 
 # In order to drop the root user, we have to make some directories world
 # writeable as OpenShift default security model is to run the container under
--- a/1.20/Dockerfile
+++ b/1.20/Dockerfile
@@ -59,10 +59,10 @@ RUN yum install -y yum-utils gettext hostname && \
     yum -y clean all --enablerepo='*'
 
 # Copy the S2I scripts from the specific language image to $STI_SCRIPTS_PATH
-COPY $NGINX_VERSION/s2i/bin/ $STI_SCRIPTS_PATH
+COPY 1.20/s2i/bin/ $STI_SCRIPTS_PATH
 
 # Copy extra files to the image.
-COPY $NGINX_VERSION/root/ /
+COPY 1.20/root/ /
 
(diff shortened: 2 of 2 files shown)


zmiklank:
[test-all]

test/run
```diff
@@ -127,9 +127,10 @@ test_docker_run_usage() {
 test_scl_usage() {
   local run_cmd="$1"
   local expected="$2"
+  local cid_file="$CID_FILE_DIR/$(mktemp -u -p . --suffix .cid)"
 
   info "Testing the image SCL enable"
-  out=$(docker run --rm ${IMAGE_NAME} /bin/bash -c "${run_cmd} 2>&1")
+  out=$(docker run --cidfile ${cid_file} ${IMAGE_NAME} /bin/bash -c "${run_cmd} 2>&1")
```
phracek, line 133: option --rm should be there. As soon as container is exited, then container can be deleted.

test/run
```diff
@@ -423,11 +424,11 @@ function run_dockerfiles_test() {
   rm "$dockerfile_tmp"
 }
 
+set -x
```
phracek, line 427: Any reason for this?

1.18/Dockerfile
```diff
@@ -59,10 +59,10 @@ RUN yum install -y yum-utils gettext hostname && \
     yum -y clean all --enablerepo='*'
 
 # Copy the S2I scripts from the specific language image to $STI_SCRIPTS_PATH
-COPY $NGINX_VERSION/s2i/bin/ $STI_SCRIPTS_PATH
+COPY 1.18/s2i/bin/ $STI_SCRIPTS_PATH
 
 # Copy extra files to the image.
-COPY $NGINX_VERSION/root/ /
+COPY 1.18/root/ /
```
phracek, line 65: We have to find out, why NGINX_VERSION variable can not be used. I would not hardcoded.

phracek (changes requested):
Thanks for this PR. I don't think, hardcoded version is proper way.

test/run
zmiklank: Oh sorry, I thought I removed this commit.

1.18/Dockerfile
zmiklank: Kind of agree. The PR can be a partial step in fixing this.

zmiklank:
[test-all]

zmiklank:
Plus another thought: this can be a bug in podman, as the podman version in centos7 compose could not be any new.
This is not happening when running tests for centos7 on fedora compose.

1.18/Dockerfile
zmiklank: It was agreed, that using the hardcoded version in here is correct option.

PR #235, opened 2023-01-13
use sclorg-specific wrapper for tfaga

zmiklank:
This PR proposes extracting duplicate code from the testing GitHub Actions in containers repositories and moving it to its own repository.
This new repository is not yet created and I would like to know what @phracek thinks about it before creating it.
The new repository would contain composite GitHub Action which will work as a wrapper around tfaga for sclorg use-cases.

This PR also shows what changes would need to be done in all of the container repositories in order to propagate this change.

It could seem like it would be better to have this new composite action in container-common-scripts repository. However, the composite action filename has to be `action.yml` and that is too non-descriptive to be included in the container-common-scripts repository.
We can try symlinks, but I personally haven't tried if that works.
Because of this I think better solution would be to create whole new repository as I stated above.

So, @phracek WDYT?

This PR is more like a discussion starter, not ready to be merged. Therefore I am creating it as a draft.

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -4,112 +4,24 @@ on:
       - created
 jobs:
   container-tests:
-    # This job only runs for '[test]' pull request comments by owner, member
     runs-on: ubuntu-20.04
-    name: "Container-tests: ${{ matrix.version }} - ${{ matrix.context }}"
+    name: "Container-tests: ${{ matrix.version }} - ${{ matrix.os_test }}"
     strategy:
       fail-fast: false
       matrix:
         version: [ "1.20", "1.22", "1.22-micro" ]
-        os_test: [ "fedora", "centos7", "rhel7", "rhel8", "rhel9", "c9s"]
-        include:
-          - tmt_plan: "fedora"
-            os_test: "fedora"
-            context: "Fedora"
-            compose: "Fedora-latest"
-            api_key: "TF_PUBLIC_API_KEY"
-            branch: "main"
-            tmt_repo: "https://github.com/sclorg/sclorg-testing-farm"
-          - tmt_plan: "centos7"
-            os_test: "centos7"
-            context: "CentOS7"
-            compose: "CentOS-7"
-            api_key: "TF_PUBLIC_API_KEY"
-            branch: "main"
-            tmt_repo: "https://github.com/sclorg/sclorg-testing-farm"
-          - tmt_plan: "rhel7-docker"
-            os_test: "rhel7"
-            context: "RHEL7"
-            compose: "RHEL-7-LatestUpdated"
-            api_key: "TF_INTERNAL_API_KEY"
-            branch: "master"
-            tmt_repo: "https://gitlab.cee.redhat.com/platform-eng-core-services/sclorg-tmt-plans"
-            tf_scope: "private"
-          - tmt_plan: "rhel8-docker"
-            os_test: "rhel8"
-            context: "RHEL8"
-            compose: "RHEL-8.6.0-Nightly"
-            api_key: "TF_INTERNAL_API_KEY"
-            branch: "master"
-            tmt_repo: "https://gitlab.cee.redhat.com/platform-eng-core-services/sclorg-tmt-plans"
-            tf_scope: "private"
-          - tmt_plan: "rhel9-docker"
-            os_test: "rhel9"
-            context: "RHEL9"
-            compose: "RHEL-9.1.0-Nightly"
-            api_key: "TF_INTERNAL_API_KEY"
-            branch: "master"
-            tmt_repo: "https://gitlab.cee.redhat.com/platform-eng-core-services/sclorg-tmt-plans"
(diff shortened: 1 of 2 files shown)


zmiklank:
I just learned, that it should be possible to create a directory in a repository (in this case in container-common-scripts) and put `action.yml` there. That would solve our naming problem an we would not need to create a whole new repository.
I will prepare PR for that.

zmiklank:
a note:

first usecase when this will we useful would be when removing deprecated "set-output" from the container's github actions, because this we need to do anyway.

zmiklank:
After discussion within the team it was decided, that better solution for this use case would be to create new repository for this wrapper.
The repository would be called `tfaga-wrapper` o `tfaga-handler`.

.github/workflows/container-tests.yml
```diff
@@ -4,112 +4,25 @@ on:
...
-          git_ref: ${{ matrix.branch }}
-          tf_scope: ${{ matrix.tf_scope }}
-          tmt_plan_regex: ${{ matrix.tmt_plan }}
-          pull_request_status_name: "${{ matrix.context }} - ${{ matrix.version }}"
-          variables: "REPO_URL=$GITHUB_SERVER_URL/$GITHUB_REPOSITORY;REPO_NAME=$GITHUB_REPOSITORY;PR_NUMBER=${{ github.event.issue.number }};SINGLE_VERSION=${{ matrix.version }};OS=${{ matrix.os_test }};TEST_NAME=test"
-          compose: ${{ matrix.compose }}
+          os_test: ${{ matrix.os_test }}
+          test_case: ${{ matrix.test_case }}
+          version: ${{ matrix.version }}
+          private_api_key: ${{ secrets.TF_INTERNAL_API_KEY }}
```
phracek, line 27: This is not needed. It can be done in `tfgag-wrapper` based on the OS host.
Secrets are available over all https://github.com/sclorg organization.

.github/workflows/openshift-tests.yml
```diff
@@ -4,103 +4,25 @@ on:
...
-            os_test: "rhel9"
-            context: "RHEL9 - OpenShift 4"
-            compose: "RHEL-9.1.0-Nightly"
-            test_name: "test-openshift-4"
-            api_key: "TF_INTERNAL_API_KEY"
-            branch: "master"
-            tmt_repo: "https://gitlab.cee.redhat.com/platform-eng-core-services/sclorg-tmt-plans"
-            tf_scope: "private"
+        os_test: [ "centos7", "rhel7", "rhel8", "rhel9" ]
+        test_case: [ "openshift" ]
```
phracek, line 14: There has to be `openshift-4` as well.

.github/workflows/openshift-tests.yml
```diff
@@ -4,103 +4,25 @@ on:
...
-          git_ref: ${{ matrix.branch }}
-          tf_scope: ${{ matrix.tf_scope }}
-          tmt_plan_regex: ${{ matrix.tmt_plan }}
-          pull_request_status_name: "${{ matrix.context }} - ${{ matrix.version }}"
-          variables: "REPO_URL=$GITHUB_SERVER_URL/$GITHUB_REPOSITORY;REPO_NAME=$GITHUB_REPOSITORY;PR_NUMBER=${{ github.event.issue.number }};SINGLE_VERSION=${{ matrix.version }};OS=${{ matrix.os_test }};TEST_NAME=${{ matrix.test_name }}"
-          compose: ${{ matrix.compose }}
+          os_test: ${{ matrix.os_test }}
+          version: ${{ matrix.version }}
+          test_case: ${{ matrix.test_case }}
+          public_api_key: ${{ secrets.TF_PUBLIC_API_KEY }}
```
phracek, line 27: the same as for container-tests.yaml. API_KEY can be defined in `tfaga-wrapper`

phracek (changes requested):

zmiklank:
Change requests addressed in [593729e](https://github.com/sclorg/nginx-container/pull/235/commits/593729ea6f025487f598ef53bdfb20ef872f01cc).

However, let me test it first in my fork.

zmiklank:
After testing and searching for answers I come into the conclusion that inherit secrets in our use case is not possible. 

Composite action by default does not support using secrets inside of them because of security.
This can be workarounded by using `secrets: inherit` in the callee's YAML file.
However, using `secrets: inherit` is for some reason only available for composite actions located in `.github/workflows/` directory. Our composite action is located directly in root of the repository; passwords inheritance is thus not working for our setup.

Only option we have is to pass secrets via inputs.

zmiklank:
@phracek this is now ready for review.

phracek (changes requested):
Please look at my comments.

.github/workflows/openshift-tests.yml
zmiklank: It can not. Please see https://github.com/sclorg/nginx-container/pull/235#issuecomment-1429833400.

phracek (approved):

PR #238, opened 2023-01-20
Update README.md

dirkrydvan:
Example for s2i has an issue with chown. It works with an additional = "ADD --chown=1001:0 app-src /tmp/src"


phracek (changes requested):
Can you please update also the rest versions?

phracek (approved):

Issue #242, opened 2023-03-20
Use STOPSIGNAL SIGQUIT instead of SIGTERM


danielkauf:
Looking at the Dockerfiles I don't find that STOPSIGNAL is set as in the official Dockerfiles for nginx (https://github.com/nginxinc/docker-nginx/blob/master/Dockerfile-debian.template#L106). Please see:
https://github.com/nginxinc/docker-nginx/issues/377

zmiklank:
Hi, thank you for creating the issue.

The difference is graceful shutdown (SIGQUIT) vs. fast shutdown (SIGTERM, SIGINT).[1]
It seems, that fast shutdown immediately closes all open connections.
However, (graceful) shutdown has 10 seconds, before docker/podman sends a SIGKILL.

Had the fast shutdown caused you any problems?

@uhliarik  what do you think about this?

[1] http://nginx.org/en/docs/control.html

danielkauf:
From what I understand, the 10 seconds for docker and 30 seconds for Kubernetes will wait are only if the container doesn't finish exit before that time. If it hasn't exited it will send a SIGKILL to terminate immediately. (kuberentes: https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace, Docker: https://vsupalov.com/docker-compose-stop-slow/). I did't have the problem yet, but the app is not yet in production, I was just creating a Dockerfile for openresty based on the nginx and notice that openresty and nginx oficial Dockerfiles had "STOPSIGNAL SIGQUIT" but this one didn't have it.

zmiklank:
We now use the SIGQUIT too, thanks again.

PR #243, opened 2023-03-29
use SIGQUIT as a STOPSIGNAL in all images

zmiklank:
Nginx upstream [1] uses SIGQUIT for graceful shutdown even in their container images [2].

We switch to SIGQUIT too, as current default SIGTERM immediately closes opened connections.

[1] http://nginx.org/en/docs/control.html
[2] https://github.com/nginxinc/docker-nginx/issues/377

Resolves: https://github.com/sclorg/nginx-container/issues/242

### DIFF
--- a/1.18/Dockerfile
+++ b/1.18/Dockerfile
@@ -90,6 +90,8 @@ RUN sed -i -f ${NGINX_APP_ROOT}/nginxconf.sed ${NGINX_CONF_PATH} && \
 
 USER 1001
 
+STOPSIGNAL SIGQUIT
+
 # Not using VOLUME statement since it's not working in OpenShift Online:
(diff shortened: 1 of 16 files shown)


zmiklank:
[test-all]

hhorak:
LGTM from my end, we might even add a link to upstream info about this in the Dockefile (as a comment, although we should be able to get reasoning and background for this from git blame as well, as the commit message mentions both).

Anyway, unless @uhliarik or @notroj gave their +1 elsewhere, I think we should get their explicit opinion first.

uhliarik:
+1 from me, as it makes sense to shutdown nginx gracefully without shutting down active connections forcibly. Thanks.

PR #246, opened 2023-05-22
Support testing nginx-container 1.24

phracek:
This pull request adds support for testing nginx-container with version 1.24.

The part of this pull request is also directory examples/1.24 which is needed for test https://github.com/sclorg/nginx-container/blob/master/test/run#L418

It clones https://github.com/sclorg/nginx-container where should be a directory 1.24.


zmiklank (commented):
Couldn't the example files be just symlinks to equivalent files that are already present?

zmiklank (approved):

PR #248, opened 2023-06-08
use dependabot for periodically updating common/

zmiklank:
This PR adds configuration for dependabot which will once a week check state of common/ and create PR with updated common/ if that is needed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,9 @@ updates:
       interval: weekly
     labels:
       - 'git-submodule'
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
+    labels:
+      - 'github-actions'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+---
+
+version: 2
+updates:
+  - package-ecosystem: gitsubmodule
+    directory: /
+    schedule:
+      interval: weekly
+    labels:
+      - 'git-submodule'


zmiklank:
@jamacku could you please review?

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+
+version: 1
+updates:
+  - package-ecosystem: gitsubmodule
+    directory: /
+    schedule:
+      interval: weekly
+    labels:
+      - 'merge-strategy: rebase'
+      - 'type: dependencies'
```
jamacku, line 11: FYI: `merge-strategy: rebase` and `type: dependencies` are just labels. Not sure if you want to use them.

I have additional automation that uses these two labels, generates changelog/release notes, and automatically merges PRs.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+version: 1
```
jamacku, line 3: There is already version `2`.

jamacku (commented):
Looks good, just two inline notes.

Since you are using GitHub Actions, did you considered also adding:

```yml
- package-ecosystem: github-actions
    directory: /
    schedule:
      interval: monthly
    labels:
      - 'github-actions'
```

zmiklank:
@jamacku Thank you for review. All the issues are now addressed in new commits.

jamacku (approved):
Looks great! :100:

zmiklank:
@phracek When this PR is merged, the dependabot will create a PR with updated common/ submodule if that is needed. (And once a month PR with bumps of github actions if available). We talked about it in the container meeting. Are you OK with merging?

PR #252, opened 2023-06-09
remove labels from dependabot.yml, as we do not use them anywhere

zmiklank:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,10 +7,10 @@ updates:
     schedule:
       interval: weekly
     labels:
-      - 'git-submodule'
+      - 'ready for review'
   - package-ecosystem: github-actions
     directory: /
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,11 +6,7 @@ updates:
     directory: /
     schedule:
       interval: weekly
-    labels:
-      - 'git-submodule'
```
jamacku, line 10: Maybe set the `ready for review` label?

jamacku (commented):
Just so you know, dependabot will set default labels even when you merge this change.

If you don't want any labels to be set by dependabot, you have to use `labels: []` - from [the documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#labels).

PR #256, opened 2023-07-10
Remove CentOS7 based images

zmiklank:
This PR also removes not needed cccp.yml and removes files from deprecated 1.18 nginx version.


zmiklank:
[test-all]

zmiklank:
[test-all]

pkubatrh:
Im a bit confused here - did we not decide to only stop testing and pushing the images to quay.io? I would rather not remove all the sources before Centos 7 EOL.

zmiklank:
> Im a bit confused here - did we not decide to only stop testing and pushing the images to quay.io? I would rather not remove all the sources before Centos 7 EOL.

Sure. I will add .exclude- files then and return the containerfiles back.

zmiklank:
[test-all]

1.20/README.md
```diff
@@ -1,9 +1,8 @@
 Nginx 1.20 server and a reverse proxy server container image
 ============================================================
 This container image includes Nginx 1.20 server and a reverse server for OpenShift and general usage.
-Users can choose between RHEL, CentOS, CentOS Stream and Fedora based images.
+Users can choose between RHEL, CentOS Stream and Fedora based images.
 The RHEL images are available in the [Red Hat Container Catalog](https://access.redhat.com/containers/),
-the CentOS images are available in the [Quay.io](https://quay.io/organization/centos7),
```
pkubatrh, line 6: I would keep these two lines - the images are still available to download and we do expect them to be used until the EOL of centos7.

pkubatrh:
otherwise lgtm

PR #261, opened 2023-08-16
stop building usupported Fedora images + move nginx 1.24 Fedora image to f38 base

zmiklank:
The "1.22/Dockerfile.fedora" and "1.22-micro/Dockerfile.fedora" has already their .exclude-files, so are not being tested. This PR disables their building and pushing to quay.io too.


zmiklank:
[test]

phracek:
@zmiklank What about adding building 1.24 and pushing to quay.io? Separate PR maybe?

zmiklank:
> @zmiklank What about adding building 1.24 and pushing to quay.io? Separate PR maybe?

Let's talk about it in the meeting. And yes, if adding, then in different PR. Thanks for the review.

PR #265, opened 2023-10-16
Get rid off obsoleted exclude files

phracek:
Get rid off obsoleted exclude files

Deleted in `1.22-micro` version:
`.exclude-rhel7` - `Dockerfile.rhel7` does not exist

Deleted in `1.22` version:
`.exclude-rhel7` - `Dockerfile.rhel7` does not exist

Deleted in `1.24` version:
`.exclude-rhel7` - `Dockerfile.rhel7` does not exist


phracek:
[test]

zmiklank (changes requested):
I think we can remove 1.22/.exclude-cento7 too.

zmiklank (approved):

PR #266, opened 2023-11-15
Move openshift tests to test-lib-nginx.sh and use it in OpenShift 4 as well

phracek:
Move openshift tests to test-lib-nginx.sh and use it in OpenShift 4 as well
Call all functions as in OpenShift 3 as in OpenShift 4


phracek:
[test-openshift]

zmiklank (approved):

PR #273, opened 2023-12-05
Version 1.22/test directory did not contain symlinks like the others

phracek:
Version 1.22/test directory did not contain symlinks like the others
This pull request fixes symlinks in 1.22/test directory


phracek:
[test-all]

zmiklank:
Hm, this is happening way too often. Do you think we could enforce some rules/checks for this? Or maybe use distgen everywhere ;).

zmiklank (approved):

phracek:
[test-all]

phracek:
> Hm, this is happening way too often. Do you think we could enforce some rules/checks for this? Or maybe use distgen everywhere ;).

No please no dist-gen :(

PR #277, opened 2023-12-21
Nginx 1.20 reached EOL for RHEL8

phracek:
Nginx 1.20 is still supported for RHEL7.

Nginx 1.20 for RHEL8 reached EOL in November 2023.

This pull request removes the version from tests and builds for RHEL8


phracek:
[test-all]

PR #278, opened 2024-01-02
Add nginx 1.24 for rhel8 and rhel9

phracek:
This pull request allows testing nginx 1.24 for C9s, C8s, rhel8, and rhel9.

Dockerfile.rhel8 is added for version 1.24

Also, build and push is updated for C9s and C8s.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -36,13 +36,13 @@ jobs:
             quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
             quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
             tag: "c8s"
-            image_name: "nginx-124-micro-c8s"
+            image_name: "nginx-124-c8s"
           - dockerfile: "1.24/Dockerfile.c9s"
             registry_namespace: "sclorg"
             quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
             quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
             tag: "c9s"
-            image_name: "nginx-124-micro-c9s"
+            image_name: "nginx-124-c9s"
 
     steps:
       - name: Build and push to quay.io registry
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -31,6 +31,18 @@ jobs:
             quayio_token: "QUAY_IMAGE_FEDORA_BUILDER_TOKEN"
             tag: "fedora"
             image_name: "nginx-124"
+          - dockerfile: "1.24/Dockerfile.c8s"
+            registry_namespace: "sclorg"
+            quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
+            quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
+            tag: "c8s"
+            image_name: "nginx-124-micro-c8s"
+          - dockerfile: "1.24/Dockerfile.c9s"
+            registry_namespace: "sclorg"
+            quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
+            quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
+            tag: "c9s"
+            image_name: "nginx-124-micro-c9s"
 
     steps:
       - name: Build and push to quay.io registry
--- a/1.24/Dockerfile.c8s
+++ b/1.24/Dockerfile.c8s
@@ -85,9 +85,4 @@ USER 1001
 
 STOPSIGNAL SIGQUIT
 
-# Not using VOLUME statement since it's not working in OpenShift Online:
-# https://github.com/sclorg/httpd-container/issues/30
-# VOLUME ["/usr/share/nginx/html"]
-# VOLUME ["/var/log/nginx/"]
-
 CMD $STI_SCRIPTS_PATH/usage
--- a/1.24/Dockerfile.c8s
+++ b/1.24/Dockerfile.c8s
@@ -1,4 +1,4 @@
-FROM quay.io/sclorg/s2i-base-c8s:c8s
+FROM quay.io/sclorg/s2i-core-c8s:c8s
 
 EXPOSE 8080
 EXPOSE 8443
--- a/1.24/Dockerfile.c8s
+++ b/1.24/Dockerfile.c8s
@@ -38,7 +38,6 @@ ENV NGINX_CONFIGURATION_PATH=${APP_ROOT}/etc/nginx.d \
     NGINX_LOG_PATH=/var/log/nginx \
     NGINX_PERL_MODULE_PATH=${APP_ROOT}/etc/perl
 
-# Modules does not exist
 RUN yum -y module enable nginx:$NGINX_VERSION && \
     INSTALL_PKGS="nss_wrapper bind-utils gettext hostname nginx nginx-mod-stream nginx-mod-http-perl" && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
--- a/1.24/Dockerfile.c8s
+++ b/1.24/Dockerfile.c8s
@@ -1,0 +1,94 @@
+FROM quay.io/sclorg/s2i-base-c8s:c8s
+
+EXPOSE 8080
+EXPOSE 8443
+
+ENV NAME=nginx \
+    NGINX_VERSION=1.24 \
+    NGINX_SHORT_VER=124 \
+    VERSION=0
+
+ENV SUMMARY="Platform for running nginx $NGINX_VERSION or building nginx-based application" \
+    DESCRIPTION="Nginx is a web server and a reverse proxy server for HTTP, SMTP, POP3 and IMAP \
+protocols, with a strong focus on high concurrency, performance and low memory usage. The container \
+image provides a containerized packaging of the nginx $NGINX_VERSION daemon. The image can be used \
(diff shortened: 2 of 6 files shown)


phracek:
[test]

.github/workflows/build-and-push.yml
```diff
@@ -31,6 +31,18 @@ jobs:
...
+            quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
+            quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
+            tag: "c8s"
+            image_name: "nginx-124-micro-c8s"
+          - dockerfile: "1.24/Dockerfile.c9s"
+            registry_namespace: "sclorg"
+            quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
+            quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
+            tag: "c9s"
+            image_name: "nginx-124-micro-c9s"
```
zmiklank, line 45:
```suggestion
            image_name: "nginx-124-c9s"
```

.github/workflows/build-and-push.yml
```diff
@@ -31,6 +31,18 @@ jobs:
             quayio_token: "QUAY_IMAGE_FEDORA_BUILDER_TOKEN"
             tag: "fedora"
             image_name: "nginx-124"
+          - dockerfile: "1.24/Dockerfile.c8s"
+            registry_namespace: "sclorg"
+            quayio_username: "QUAY_IMAGE_SCLORG_BUILDER_USERNAME"
+            quayio_token: "QUAY_IMAGE_SCLORG_BUILDER_TOKEN"
+            tag: "c8s"
+            image_name: "nginx-124-micro-c8s"
```
zmiklank, line 39:
```suggestion
            image_name: "nginx-124-c8s"
```

1.24/Dockerfile.c8s
```diff
@@ -0,0 +1,94 @@
+FROM quay.io/sclorg/s2i-base-c8s:c8s
```
zmiklank, line 1: Is there any reason for building Centos Stream images on top of s2i-base instead of s2i-core?
I doublechecked and this issue is actual also for 1.20/Dockerfile.c9s and 1.22/Dockerfile.c9s.

phracek: Good catch, fixed by https://github.com/sclorg/nginx-container/pull/278/commits/e677713face138b5810d00284e03585a99f1d657

zmiklank: Thanks, could you please also fix these two: 1.20/Dockerfile.c9s and 1.22/Dockerfile.c9s? So we do not forget about it. Or should we leave them as they are, if they are already released?

phracek:
[test]

phracek:
[test-openshift]

phracek:
[test-all]

.github/workflows/build-and-push.yml
hhorak, line 45: The `micro` part looks suspicious -- this is not a micro one, is it?

hhorak: Ha, now I see @zmiklank already suggests a change..

hhorak:
LGTM, except the `image_name` in the build-and-push, that @zmiklank spotted.

hhorak:
Hups, accidental closure O:-D

1.24/Dockerfile.c8s
```diff
@@ -0,0 +1,94 @@
...
+
+ENV NGINX_CONFIGURATION_PATH=${APP_ROOT}/etc/nginx.d \
+    NGINX_CONF_PATH=/etc/nginx/nginx.conf \
+    NGINX_DEFAULT_CONF_PATH=${APP_ROOT}/etc/nginx.default.d \
+    NGINX_CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/nginx \
+    NGINX_APP_ROOT=${APP_ROOT} \
+    NGINX_LOG_PATH=/var/log/nginx \
+    NGINX_PERL_MODULE_PATH=${APP_ROOT}/etc/perl
+
+# Modules does not exist
```
zmiklank, line 41: What this comment means?

1.24/Dockerfile.c8s
```diff
@@ -0,0 +1,94 @@
...
+    rpm-file-permissions
+
+USER 1001
+
+STOPSIGNAL SIGQUIT
+
+# Not using VOLUME statement since it's not working in OpenShift Online:
+# https://github.com/sclorg/httpd-container/issues/30
+# VOLUME ["/usr/share/nginx/html"]
+# VOLUME ["/var/log/nginx/"]
```
zmiklank, line 92: Is this still actual?

.github/workflows/build-and-push.yml
zmiklank: I suggested the change long ago, meanwhile marked as resolved. I have unresolved it now.

zmiklank: Also I see that 1.24/Dockerfile.c9s has .exclude-file. So this will not work.

phracek: Fix https://github.com/sclorg/nginx-container/pull/278/commits/67ecede3d2da94df21b4fb2c5b0e5fe0349ecdd7

phracek: It was delete ;)

.github/workflows/build-and-push.yml
phracek: Fixed by https://github.com/sclorg/nginx-container/pull/278/commits/67ecede3d2da94df21b4fb2c5b0e5fe0349ecdd7

.github/workflows/build-and-push.yml
phracek: Fixed by https://github.com/sclorg/nginx-container/pull/278/commits/67ecede3d2da94df21b4fb2c5b0e5fe0349ecdd7

1.24/Dockerfile.c8s
phracek: Fixed by https://github.com/sclorg/nginx-container/pull/278/commits/67ecede3d2da94df21b4fb2c5b0e5fe0349ecdd7. Copy from 1.22/Dockerfile.c8s ;)

1.24/Dockerfile.c8s
phracek: I don't know, but VOLUMES are generally working.

.github/workflows/build-and-push.yml
zmiklank: Peto, please, do not close issues that are not resolved. There is a higher change they would pass the review unfixed then.

Are you sure we want to have build and push enabled for nginx-124-c9s when the .exclude-file is present?
What are the reasons for this?

zmiklank (changes requested):
Please see my comments.

1.24/Dockerfile.c8s
phracek: Based on the response from issue mentioned in the comment. We can remove it. I will do it.

phracek: This issue https://github.com/sclorg/httpd-container/issues/30 was closed as it is not an issue anymore on OpenShift 4.
This commit https://github.com/sclorg/nginx-container/pull/278/commits/5d9f130194081cf8cdafcee73e44e075d4d7083a removes comments

.github/workflows/build-and-push.yml
phracek: @zmiklank The file is not there. It was deleted. And as this is not UBI image, we usually want to provide a C9S image to quay.io.

1.24/Dockerfile.c8s
```diff
@@ -85,9 +85,4 @@ USER 1001
 
 STOPSIGNAL SIGQUIT
 
-# Not using VOLUME statement since it's not working in OpenShift Online:
```
zmiklank: If I understand this correctly, the issue with volumes is no longer present on OCP.
In that case, don't we want to use the VOLUMEs in Containerfiles instead of removing them entirely?

.github/workflows/build-and-push.yml
phracek: And finally, everything was solved.

1.24/Dockerfile.c8s
phracek: @bparees Can you please respond more clearly to why we do not need to define VOLUME directive, I guess I explained it but you will be more precise. Thank you

bparees: They were really never needed.  VOLUME defines implicit volume mounts, but it's better to just use explicit ones (meaning either using "-v" when running an image via docker, or using a volume/volumemount in the pod definition in k8s/openshift).  So you could always mount a filesystem to a path in a container, even if the image didn't explicitly define a VOLUME. 

the problem was that when docker saw an image with a VOLUME definition it would automatically try to mount a volume(empty) into that path, which openshift online did not allow.  But i do not believe that behavior is part of OCP any more with the move off of docker as the container runtime.

It is somewhat similar to the "EXPOSE" declaration....it's more of a hint than a necessary declaration.  The difference is EXPOSE does literally nothing at runtime, whereas VOLUME does cause a volume to be mounted at least with docker's container runtime.

some more background: https://stackoverflow.com/questions/40163036/difference-between-volume-declaration-in-dockerfile-and-v-as-docker-run-paramet

bparees: the other thing that changed is how openshift runs s2i builds. It no longer talks to the docker socket to launch a container running the s2i builder image.  So it's no longer relevant what VOLUMES the s2i builder image declares, so you won't hit this issue any more even if the image does declare a VOLUME.

zmiklank (approved):

PR #282, opened 2024-03-06
${{ github.head_ref }} not available for issue comment trigger

zmiklank:
We can use group based on issue number - so that on every PR only one parallel tests for containers and ocp could run.

Using ${{ github.head_ref }} is not valid for our use-case, see https://github.com/sclorg/nginx-container/actions/runs/8158453585

@jamacku PTAL.


jamacku (approved):
make sense :+1:, LGTM

PR #283, opened 2024-03-06
tests: use per job parallel group

zmiklank:
@jamacku PTAL.

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -3,14 +3,13 @@ on:
     types:
       - created
 
-concurrency:
-  group: container_tests-${{ github.event.issue.number }}
-  cancel-in-progress: true
-
 jobs:
   container-tests:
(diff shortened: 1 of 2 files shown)


.github/workflows/container-tests.yml
```diff
@@ -3,13 +3,12 @@ on:
...
-  group: container_tests-${{ github.event.issue.number }}
-  cancel-in-progress: true
-
 jobs:
   container-tests:
     runs-on: ubuntu-20.04
+    concurrency:
+      group: container-${{ github.event.issue.number }}-${{ matrix.version }}-${{ matrix.os_test }}
+      cancel-in-progress: true
     name: "Container-tests: ${{ matrix.version }} - ${{ matrix.os_test }}"
```
jamacku, line 12: You could also move `name` to the beginning of the job, so it is consistent across the workflows.

jamacku (approved):
LGTM

PR #290, opened 2024-03-25
Add testing NginX 1.24 container on OpenShift 4

phracek:
This pull request adds testing nginx 1.24 container on OpenShift 4.

Container tests are already tested.


phracek:
Tests are not needed to execute. It modifies only GitHub Action.

PR #299, opened 2024-05-21
deprecate c8s images, as EOL for c8s is 31th of May 2024

zmiklank:
Wait until 31th of May for merge.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -38,14 +38,6 @@ jobs:
             image_name: "nginx-124"
 
           - version: "1.24"
-            tag: "c8s"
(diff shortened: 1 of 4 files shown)


phracek (changes requested):
In case we delete CentOS 7 based images from READMEs, I would prefer to replace them with C9S.

And I guess, we can replace *RHEL7 based image* with RHEL8 or even RHEL9

zmiklank:
> In case we delete CentOS 7 based images from READMEs, I would prefer to replace them with C9S.

You are right, I have removed changes regarding CentOS7 from this PR. They will be addressed later.

phracek (approved):
LGTM. Thanks

PR #305, opened 2024-06-20
Add 1.26 examples directory

phracek:
This is needed for passing tests https://github.com/sclorg/nginx-container/pull/297

We have a test run_dockerfiles_test that clones upstream repository and adds examples/<version> application.
But in case of new version it does not exist yet ;)

This pull request avoids failing tests.


phracek:
This pull request is mandatory for #297 
[test]

phracek:
Fedora 1.24 failure is fixed by #297 as well CentOS Stream 9 - 1.22-micro

PR #310, opened 2024-07-17
Use Python 3.12 for pytest and option -rA

phracek:
Let's use Python3.12 as we decided globally.


phracek:
[test-openshift-pytest]

PR #314, opened 2024-07-29
Allow testing nginx-container on C10S host

phracek:


phracek:
Running tests is not needed. The change is relevant only for GitHub Action.

PR #316, opened 2024-09-17
Remove `run-openshift` that are deprecated. Adds helm charts tests

phracek:
Adds helm charts tests


phracek:
[test-openshift-pytest]

PR #317, opened 2024-09-18
This pull request replaces nss_wrapper with nss_wrapper-libs

phracek:
This pull request replaces nss_wrapper with nss_wrapper-libs

Less dependency and unnecessary footprint


phracek:
[test]

PR #321, opened 2024-11-06
Fix helm charts order

phracek:
This pull request does not modify source neither tests. Only change order of test execution.

Container-ci-suite do no support switching between clusters.


phracek:
[test-all]

phracek:
No code is changed. Let's get merge it. Everything PASSED.

PR #327, opened 2025-02-13
Fix nginx-container OpenShift 4 tests

phracek:


phracek:
[test-openshift-4][test-openshift-pytest]

PR #328, opened 2025-03-04
Support testing nginx-container on shared_cluster

phracek:
This pull request enables testing nginx-container on shared cluster

The nginx-container code is not changed at all.


phracek:
[test-all]

PR #326, opened 2025-01-21
Allow building and testing Nginx-1.26 on RHEL9 and C9S

phracek:
Remove exclude files. Allow building nginx-126 on RHEL9 and C9S

PR #330, opened 2025-03-21
Replace 'bind-utils' with 'bind9.18-utils'

phracek:
Replace 'bind-utils' with 'bind9.18-utils'

See Jira issue https://issues.redhat.com/browse/RHEL-80345

It is valid only for RHEL9


SlouchyButton (commented):
Why the switch from smaller set of modules to `nginx-all-modules`, since it's not mentioned in PR nor in commit message.

Is it an unrelated change just 'hidden' in the PR, or mistake, or required for the `bind-utils` change?

phracek:
> Why the switch from smaller set of modules to `nginx-all-modules`, since it's not mentioned in PR nor in commit message.
> 
> Is it an unrelated change just 'hidden' in the PR, or mistake, or required for the `bind-utils` change?

@SlouchyButton See PR here https://github.com/sclorg/nginx-container/pull/326. I did not want to create a separate PR. But I will definitely add a comment into commit message. I have missed it. Good point.

PR #331, opened 2025-04-03
For S2I builds do not use shared cluster

phracek:
For S2I builds do not use shared cluster
It has an issue with pulling image like ErrImagePull.

Needs more investigation.


phracek:
[test-openshift][test-openshift-pytest]

phracek:
Tests have passed. No container code is changed. let's get merge it so we can see if nighly builds are passing.

PR #334, opened 2025-05-06
Do not test curl connection

phracek:
Do not test curl connection.

It does not make sense. We used 'helm test' that is enought.


phracek:
[test-all]

PR #338, opened 2025-06-12
Fix reg expression for getting rhel name from 'rhel10'

phracek:
Fix reg expression for getting rhel name from 'rhel10'

The new expresssion removes all digits.

### DIFF
--- a/test/test-lib-nginx.sh
+++ b/test/test-lib-nginx.sh
@@ -29,7 +29,7 @@ function test_nginx_imagestream() {
       return
     fi
   fi
-  ct_os_test_image_stream_s2i "${THISDIR}/imagestreams/nginx-${OS%[0-9]*}.json" "${IMAGE_NAME}" \
+  ct_os_test_image_stream_s2i "${THISDIR}/imagestreams/nginx-${OS//[0-9]/}.json" "${IMAGE_NAME}" \
                               "https://github.com/sclorg/nginx-container.git" \
                               "examples/${VERSION}/test-app" \
                               "Test NGINX passed"


FaramosCZ:
I understand is as you are using BASH parameter substitution.
The change looks fine it terms fixing the expression.

However in the container-files themselves, you advocated them to not rely on BASH, just SH.
Why do you want BASH here?

Is there any expectation user may want to run these tests in an environment without BASH?
If yes, what integrity do we have on where we are fine with using bash and when we are not ?

phracek:
> I understand is as you are using BASH parameter substitution. The change looks fine it terms fixing the expression.
> 
> However in the container-files themselves, you advocated them to not rely on BASH, just SH. Why do you want BASH here?
> 
> Is there any expectation user may want to run these tests in an environment without BASH? If yes, what integrity do we have on where we are fine with using bash and when we are not ?

The testing suite is called on BASH. https://github.com/sclorg/container-common-scripts/blob/master/test.sh#L1. It is not executed in container, but on host OS.

We are focussed on BASH for long time.

phracek:
> Which leads to my original point - since we assume the Bash is available for the tests, can we **also make sure** that it is available for the container production builds, so we can benefit from the Bash parameter expansion and substitution inside of the containerfiles too ?

Production builds are totally out of our scope. In case something, from build point of view does not work, then production builds have to solved. And we don't care, but as far as I know they used BASH and not SSH.
And this fix is simpler one which work in BASH at all. E.g. https://unix.stackexchange.com/questions/245362/remove-numbers-strip-numeric-characters-from-the-string-variable.
Our preference is to use BASH in all environments.

phracek:
> I don't see any integrity in restricting us to Sh for the container production builds, but employing Bash for everything else.
> Even more when all of our development and test environments have Bash, ... just not the production buildsystem environment.

Our CI builds used only BASH and not SH. This is a prerequisite for our CI systems.

phracek:
> > Our preference is to use BASH in all environments.
> > Great, I'm all for it!
> 
> Can we make sure it's available in the production build environment too, to avoid issues like these: [sclorg/mariadb-container@486ae3f](https://github.com/sclorg/mariadb-container/commit/486ae3fa5c3872ac4beaecdb9350c66c8290d4c8) ?

Yes, please file a ticket to Konflux team, which is responsible for builds.

FaramosCZ:
Great! Will do.

Can you @phracek fix it on the upstream in the meainwhile?
On the `buildah` running in GitHub actions on the Ubuntu runner for the creation of Fedora images?

FaramosCZ:
I filed the ticket for Konflux.

Why does it make sense to you to fix it in Konflux, but not on the Fedora upstream?

Why would you want the Fedora images differ in this way?
Both inside of the container files (once the ticket is finished we will be able to use it in RHEL and CentOS Stream, but not in Fedora) and outside (production build environment for Fedora differ from our development and testing environment) ?

FaramosCZ (approved):

PR #339, opened 2025-06-23
Allow testing nginx-container on shared_cluster

phracek:


phracek:
[test-openshift][test-openshift-pytest]

PR #341, opened 2025-08-28
Remove RHEL7, CentOS 7 notes from README. Remove 'scl enable'

phracek:
This pull request removes RHEL7, CentOS7 notes from README's and
replaces ubi8 -> ubi9.

Also <version>/root/app-root/etc/scl_enabled is removed. As SCL are outdated.


phracek:
[test]

phracek:
Let's re-test container.

bind9.18-utils can be renamed to bind-utils the problem with installation was resolved.

[test]

phracek:
Remove rhel7 / ubi7 from comments as well.

[test-all]

PR #372, opened 2026-04-02
Make CentOS ImageStream json actually based on CentOS containers

daveol:
It seems something went wrong when creating the file `imagestreams\nginx-centos.json` since they point to the UBI images instead of the images produced based off CentOS Stream. This fixes that.

## Summary by CodeRabbit

* **Chores**
  * Switched nginx ImageStream base images from UBI to CentOS Stream and updated source image references.
  * Renamed image tags from `*-ubi*` to `*-el*` (e.g., `1.26-ubi9` → `1.26-el9`) and set the "latest" tag to `1.26-el9`.
  * Renamed distro labels to "CentOS Stream 8/9/10"; CentOS Stream 8 now lists only `1.24`.

### DIFF
--- a/imagestreams/imagestreams.yaml
+++ b/imagestreams/imagestreams.yaml
@@ -13,15 +13,15 @@
     https://github.com/sclorg/nginx-container/blob/master/APP_VERSION/README.md.
   imagestream_files:
   - filename: nginx-centos.json
-    latest: "1.26-ubi9"
+    latest: "1.26-el9"
     distros:
-      - name: UBI 8
-        app_versions: ["1.22", "1.24"]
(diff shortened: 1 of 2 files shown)


pkubatrh:
Thanks for the PR! This is however the expected state - the expectation here is to use UBI-based container images for workloads on top of Centos Stream.
UBI registries only contain components needed to build the UBI-based images after a given RHEL version gets released, so we build Centos Stream images to bridge the gap between the components (rpms/modules) being built and the components being available in the UBI registries.
We usually stop building the centos stream images after respective UBI-based images get released, but this is often forgotten.

PR #373, opened 2026-04-15
1.26 minimal container

ndavidova:
I was testing this by running:
```
 cp 1.26/Dockerfile.11 1.26/Dockerfile.10
 make build TARGET=rhel10 VERSIONS=1.26
```

When i built the minimized image, the size was: 109 MB, compared to previous 297MB.

### DIFF
--- a/1.26/Dockerfile.rhel11
+++ b/1.26/Dockerfile.rhel11
@@ -1,13 +1,11 @@
-FROM ubi8/ubi AS build
+FROM ubi10/ubi AS build
 
 RUN mkdir -p /mnt/rootfs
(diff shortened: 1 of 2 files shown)


phracek (changes requested):
Thanks for doing on this pull request.

Some information can you please create a totally separate directory like `1.26-minimal`, where the Dockerfile.rhel11 and the other files will be stored?

I would prefer to use not micro approach but either `ubi10` or `ubi10-minimal`, as this is a proof of concept.
You can take an inspiration here: https://github.com/sclorg/postgresql-container/pull/653

instead of `dnf` let's use `microdnf`.

PR #375, opened 2026-04-20
Compare nginx version against the major OS version

phracek:
Compare nginx version against the major OS version.

In case image is less then previous major OS version, then the test passed.

In case image is higher then previous major OS version, then the test is skipped.

## Summary by CodeRabbit

* **Tests**
  * Added container size comparison tests to verify built images are smaller than published baselines and skip appropriately for non-applicable versions.
  * Improved test documentation, reorganized parametrization and formatting for clearer, more maintainable tests.
  * Added a helper to skip tests unless running a "minimal" version.

* **Chores**
  * Cleaned up version-specific test entry points, adding and removing lightweight wrappers to align test discovery.

### DIFF
--- a/1.20/test/test_container_sizes.py
+++ b/1.20/test/test_container_sizes.py
@@ -1,0 +1,1 @@
+../../test/test_container_sizes.py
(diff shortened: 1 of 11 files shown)


phracek:
[test-pytest]

phracek:
[test-pytest]

PR #377, opened 2026-04-21
Compare sizes only for RHEL 10 and higher. No less OS

phracek:
## Summary by CodeRabbit

* **Tests**
  * Refactored container size comparison tests to skip RHEL 7, 8, and 9 versions for improved test accuracy.


phracek:
[test-pytest]

